PR #34, opened 2024-08-27
fix: consistent virus names, "Zaire Ebola Virus" and "Sudan Ebola Virus"

corneliusroemer:


theosanderson:
For me this isn't ideal formatting (I agree original insn't either). Let's discuss synchronously next meeting?

Issue #38, opened 2024-08-27
Reserve `PP_` for production


theosanderson:
We intended to reserve the `PP_` access prefix for production and to use `PPD_` eleswhere. We still need to do that.

Ultimately we should make remove `PP_` from the Values.yaml and move it to just the production yaml. But for now as a low risk change I will add `PPD_` to staging/demo-specific Values and redeploy that.

(thanks to @emmahodcroft  for raising)

PR #42, opened 2024-08-27
Add `PP_` prefix to production-specific values

theosanderson:
This adds the `PP_` prefix to production-specific values (but they are still currently duplicated in the main pathoplexus `Values.yaml`) - this is another low-risk change

relates to #38


emmahodcroft (approved):
This change makes sense to me

PR #44, opened 2024-08-27
Remove FAQ category links?

theosanderson:
Now that we have the nice collapsible FAQ sections, in my opinion one doesn't necessarily need the table of contents and to me it's a bit confusing because it has a somewhat similar bulleted formatting to the questions themselves

Old:
image

New:
https://preview-faq-category-links.pathoplexus.org/about/faq

But np if you disagree


emmahodcroft (approved):
Thanks Theo and sorry!!!

PR #52, opened 2024-08-28
fix lettering of privacy policy

theosanderson:


emmahodcroft (approved):
Looks very thorough to me!

PR #53, opened 2024-08-29
feat: add simple script-generated sitemap to /public folder

corneliusroemer:
https://preview-manual-sitemap.pathoplexus.org/sitemap.txt

Brave Browser 2024-08-29 03 38 59


theosanderson:
It looks to me like Google has already indexed many of these even in the first couple of days - I would wait to see if we need this before adding it due to the need to maintain it

corneliusroemer:
Many but possibly not all, and I don't think we would need to maintain it - there's no harm in it being not complete - but it would ensure google finds them all.
But I guess doesn't matter much anymore. Happy to close.

PR #54, opened 2024-08-29
chore(config): don't send noindex robots header for production

corneliusroemer:
Follow up to loculus-project/loculus#2657 to make production remain indexed.

This can be merged as soon as we agree on the Loculus PR, there's no harm in having an extra values.yaml that isn't used yet.

https://preview-do-index-production.pathoplexus.org


chaoran-chen:
Now that we plan to set the default to allow indexing, we don't need this anymore, right? (Or should we still set the value explicitly?)

theosanderson:
Agree that we won't need this (but thanks for making it)

corneliusroemer:
We now need the equivalent for all the pages we don't want to index ;)

PR #58, opened 2024-08-29
fix: make sure that the bug re/ unaligned sequences does not happen on pathoplexus by disabling dateToSortBy optimization

Taepper:
The recent SILO PR https://github.com/GenSpectrum/LAPIS-SILO/pull/549 contains a bugfix that was reproduced when dateToSortBy is set, to prevent this bug from happening on pathoplexus, I would disable this field before we wait for the change to be propagated upstream


theosanderson:
We don't sort on downloads - does that mean this isn't a problem?

Taepper:
No, it can also occur when no sorting is requested. It is due to internal sorting in the table sheet

chaoran-chen (approved):

PR #62, opened 2024-08-29
Improve DUT format

### LINKED ISSUE #57: [Docs] Improve formatting for Restricted-Use Data '3 points' in DUT
This is **not urgent**, but I think the formatting of our '3 points' for using Restricted-Use Data in publications and preprints in the DUT, could be improved. I had a go at this but was not very successful!
(See header just below [this section](https://pathoplexus.org/about/terms-of-use/data-use-terms#422-publications-and-preprints))

How it looks:
image

For comparison, here's how we formatted it in Google Docs before moving to the website (which I think is much clearer):
image

If we could somehow change the formatting to be a little more white-space-y and broken up, like the above, I think that would help!

theosanderson:
Resolves #57 

Should look like this:

image


emmahodcroft (approved):
This looks great - thank for that!

PR #71, opened 2024-08-31
[DO NOT MERGE] Investigate old state

chaoran-chen:
Preview: https://preview-old-state.pathoplexus.org/


corneliusroemer:
Don't we need to change the loculus commit we target?

chaoran-chen:
It is branched off from an old commit. You cannot create a PR to merge into main from a branch that does not have any novel commit.

corneliusroemer:
It did deploy, your URL was wrong :D on pathoplexus need to add `preview-` prefix (I've fixed the URL)

theosanderson:
The format of the front page does look like the old state

corneliusroemer:
My bad, yes this should be correct - I erroneously assumed the "Files changed" shows a diff versus the PR target current state. It doesn't, it shows the diff against _base_.

Fork 2024-08-31 18 12 58

PR #84, opened 2024-09-02
feat: Add robots noindex headers to all but production

corneliusroemer:
See https://github.com/loculus-project/loculus/pull/2657

https://preview-noindex-non-prod.pathoplexus.org

Also some small loculus changes are now included due to this PR updating loculus base commit https://github.com/loculus-project/loculus/compare/09241e97a6eb9bc26d7848f5bbc847b7907ba2ad...61fb6757861f47fd9919f4e0db655cdf15b17a40

image

That's what should happen:
image

### DIFF
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -2,7 +2,7 @@ accessionPrefix: "PP_"
 runDevelopmentMainDatabase: false
 runDevelopmentKeycloakDatabase: false
 bannerMessage: ""
-robotsNoindexHeader: true
+robotsNoindexHeader: false
 subdomainSeparator: "."
 secrets:
   keycloak-admin:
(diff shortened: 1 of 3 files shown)


loculus_values/environment_specific_values/production.yaml
```diff
@@ -2,6 +2,7 @@ accessionPrefix: "PP_"
 runDevelopmentMainDatabase: false
 runDevelopmentKeycloakDatabase: false
 bannerMessage: ""
+robotsNoindexHeader: true
```
corneliusroemer, line 5:
```suggestion
robotsNoindexHeader: false
```

pathoplexus_app/values.yaml
theosanderson, line 1: Once the preview comes up let's check CCHF segment downloads given that this reverts the hotfix?

corneliusroemer: I tested this on main.loculus.org already but yes, why not here as well!

corneliusroemer: So I didn't check very carefully but the output now is there - previously the downloaded file was empty.

theosanderson: Tried some CCHF dls which looked good

PR #93, opened 2024-09-03
ci: remove temporary loculus_repo directory created in action

### LINKED ISSUE #80: [BUG] Accidental git submodule
Currently this repo contains the `loculus_repo` as a git submodule. This gets added by the GitHub action I added to create PRs to update the Loculus version. It's not actually needed for that and not intended. But it also doesn't do much harm. We can fix this at some point.

corneliusroemer:
Resolves #80


theosanderson (approved):
Thanks!

theosanderson:
(For reference, if you'd wanted to you could have tested it by running the Action on branch - not saying you should have done, I'm confident-ish it will work)

corneliusroemer:
Yeah I know I can somehow test but not worth it here as it's CI only and wouldn't break anything real ever

Brave Browser 2024-09-03 15 16 37

it worked, confirmed

PR #108, opened 2024-09-05
Improvements to `sync.sh` script used for website dev

theosanderson:
Updates the `sync.sh` script used for website development.

Simplifies it by:
- not using authentication, now loculus is public
- always using the commit specified in Values.yaml, rather than doing that as a special case

Improves performance by shallow cloning just that commit. Improves error handling by setting `-e`.

### DIFF
--- a/sync.sh
+++ b/sync.sh
@@ -1,37 +1,10 @@
 #!/bin/bash
 
-helpFunction()
-{
-   echo ""
-   echo "Usage: $0 [-l] [-c commit_hash] [-a]"
-   echo -e "\t-l  Rely on system's Git credentials instead of using the LOCULUS_PAT environmental variable"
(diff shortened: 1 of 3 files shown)


sync.sh
```diff
@@ -1,37 +1,10 @@
 #!/bin/bash
 
-helpFunction()
```
corneliusroemer, line 3: I think some help would still be nice, even if it doesn't take any args

corneliusroemer (approved):
Haven't tested locally but if this breaks anything we'll notice soon and I suppose you have tested it.

Great to shallow clone etc

sync.sh
theosanderson: That adds quite a bit of complexity - we would have to maintain argparsing just for the sake of the help. I think I will leave this for now - but thanks for suggesting.

PR #123, opened 2024-09-09
chore: include file-downloads & outlinks in plausible config

corneliusroemer:
For the non-production sites we might want to set up an extra domain if we wanted to also keep track of them (for now possibly best to only use production, as otherwise we count a lot of dev traffic)


theosanderson (changes requested):
I just learnt we can track specifically by subdomain on Plausible :) 
image

https://plausible.io/docs/subdomain-hostname-filter

theosanderson:
(Downloads would be really useful though!)

corneliusroemer:
Here we go

theosanderson (approved):
Thanks!!

PR #125, opened 2024-09-12
Add EB Resolutions from 12 Sept

emmahodcroft:
Adds PDF of recent EB Resolutions (up to 12 Sept) and link on the minutes page. 

Tested locally but happy if someone wants to test again & see if all looks good.

https://preview-add-eb-res.ppx.bio

PR #126, opened 2024-09-12
Add EB Resolutions from 12 Sept #125

corneliusroemer:
Adds PDF of recent EB Resolutions (up to 12 Sept) and link on the minutes page.

Tested locally but happy if someone wants to test again & see if all looks good.

https://preview-add-eb-res.ppx.bio/

Same PR as #125 by @emmahodcroft just with branch renamed so we can have a preview (no underscores, but hyphens)


anna-parker (approved):
works for me!

PR #127, opened 2024-09-12
Add date range to EB Res PDF

emmahodcroft:
Adds date range to the PDF of the EB resolutions update for 12 Sept


monorepo/website/public/about/governance/minutes/2024-09-12-EB_Resolutions.pdf
chaoran-chen, line 1: This is not replacing the other file which is called `2024-09-12_EB_Resolutions.pdf`: with an `_` between the date and the remaining name.

emmahodcroft:
Thanks for the catch Chaoran, I'll fix this later today!

chaoran-chen (approved):

PR #139, opened 2024-09-15
feat(docs): Update metadata; include latitude, longitude allow download of all fields and descriptions in a csv file

anna-parker:
Preview:  https://preview-update-metadata-fields.pathoplexus.org/

- Add description of new latitude and longitude fields added to Loculus in https://github.com/loculus-project/loculus/pull/2772 (also add these fields to pathoplexus values.yaml) and update loculusVersion to include new metadata fields.
- Add option to download all metadata fields and descriptions in docs, delete "all optional fields" section as duplication of fields described elsewhere on the page.
- Add banner to production to tell users we might be experiencing a temporary downtime. 

image

### TO DO:

After adding these fields we need to update the pipeline version to reprocess all data and add these fields - it should not be done in the same PR so that prepro doesnt start updating without the new fields.


theosanderson:
There doesn't seem to be a change to the Loculus values.yaml to add a field. That would be required for this to actually have an effect I think. Then I'm not 100% sure if one would need to have a preprocessing version 2 or something.

corneliusroemer:
Remember not to use underscores in branch names for pathoplexus in particular at the moment :)

Any reason to use CSV and not TSV? we generally use tsvs so might be more consistent to do that here too.

corneliusroemer:
We should eventually auto generate that metadata description file from whatever is in the values yaml for a particular organism, otherwise it will quickly get out of sync.

Also, there's a chance we'll have different metadata fields for different organisms - that's allowed so we can't assume there's going to be one shared format going forward.

corneliusroemer:
The coordinates seem to be missing from:
a) the metadata CSV
b) values.yaml (as Theo has already pointed out)

We won't need a version 2 of preprocessing afaik, that's only necessary if we want to _reprocess_ already processed sequences - as we don't have coordinates annotated in any existing sequences there's no need to reprocess.

Unless the backend requires explicit null values for all existing fields, but that would be a pain and we might want to get rid of that requirement to not make metadata field additions a pain.

anna-parker:
> There doesn't seem to be a change to the Loculus values.yaml to add a field. That would be required for this to actually have an effect I think. Then I'm not 100% sure if one would need to have a preprocessing version 2 or something.

Sorry not quite sure I understand what you mean here, do you think that adding a new metadata field in https://github.com/loculus-project/loculus/pull/2772 will cause an issue in pathoplexus if we don't update the preprocessing pipeline version?

corneliusroemer:
Pathoplexus has its completely separate values yaml. It's totally uncoupled so you need to mirror the changes to the Pathopexus values yamls to have effect.

corneliusroemer:
We need to check on a persistent instance whether anything blows up when we add metadata fields without bumping the preprocessing pipeline version. Probably not but I wouldn't rely on it.

theosanderson:
> It's totally uncoupled 

This isn't strictly true - it inherits from Loculus values, and then overwrites stuff, but the standard organism config is something that would be hard to inherit even if we wanted to - because it is applied to the various organisms using yaml syntax

corneliusroemer:
Ah sorry, I thought we had totally uncoupled it. Thanks for correcting

anna-parker:
From my local testing running `helm update` after changing metadata fields has no impact on the database locally. However, I am not quite convinced I have done this correctly. I will try to redo this on a preview.

theosanderson:
My expectation would be: it would have no effect on the dB, but that the schema change might cause crashes somewhere (because an "expected" value is now undefined, if we don't tolerate that)

anna-parker:
Update: I tested on a persistent DB in https://github.com/loculus-project/loculus/pull/2807 - the db and backend are fine however SILO has unexpected behavior. Until new sequences are uploaded with these metadata fields the SILO pods are unable to start with the new configs and will not include the new fields in the downloads. As soon as new data is uploaded the SILO pods are able to restart. (I do not understand why this behavior is happening)

It does seem like we need to update the preprocessing pipeline version to do this

corneliusroemer:
Thanks @anna-parker for the testing!

Let's understand (maybe you know already, then please tell me):
- Why does SILO not start?
- What does SILO require that we're not giving it?

Based on the answer to these questions we can then consider how to proceed.

anna-parker:
@corneliusroemer I looked into this some more and its actually not to odd behavior, see https://github.com/loculus-project/loculus/pull/2807 for details. Summary: the database stores all processed metadata in a json, the backend sends this unprocessed json to silo. Silo preprocessing then checks that all metadata fields are in the data it receives.

There are in my opinion only 2 ways to resolve this: 
1. Increase the preprocessing pipeline version so that preprocessing adds the new fields to all samples.
2. Move forward with: https://github.com/loculus-project/loculus/issues/2793 - where we do not require metadata fields that are null to be sent in requests - this would mean a larger architectural change.

loculus_values/environment_specific_values/production.yaml
```diff
@@ -4,6 +4,7 @@ runDevelopmentKeycloakDatabase: false
 bannerMessage: ""
 robotsNoindexHeader: false
 subdomainSeparator: "."
+bannerMessage: "Some portions of our website are experiencing intermittent outages, and some requests may take longer than normal to complete. We apologize for the inconvenience and thank you for your patience as we resolve the issue."
```
theosanderson, line 7: This sounds like unplanned issues
```suggestion
bannerMessage: "We are rolling out a new version of the Pathoplexus processing pipeline. Sequences may take longer to process and there may be downtime of our query API. We apologise for any inconvenience."
```

theosanderson (approved):

PR #146, opened 2024-09-16
do not merge: crossref down loculus

corneliusroemer:
https://preview-crossrefdown.ppx.bio


anna-parker:
I added the new crossref PR to the loculusValues field in the same way here: https://github.com/pathoplexus/pathoplexus/pull/148

PR #151, opened 2024-09-17
ci: Remove link checking from pull requests

theosanderson:
This manifests as very flaky at the moment - partly because you just might not have a preview instance - which is a bit distracting. I think we should add this back, but once we're doing it fully within CI.

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -1,7 +1,5 @@
 name: broken-link-check
 on:
-  pull_request:
-    types: [opened, reopened, labeled, unlabeled, synchronize]
   workflow_dispatch:
     inputs:
       custom_subdomain:


corneliusroemer:
Hmm, I'm not sure I agree here. If previews are flaky we should fix that. The link checker is a preview health indicator at the very least.

If there are website that are flaky like ENA, then let's exclude them.

Where have you seen failures? I've only seen the ENA one.

corneliusroemer:
If we can stabilize it sure, but let's see first for a few more days? It takes just a click to see what failed? Also, we can just ignore/down weight failure in our minds. If we know it's flaky let's not take it too seriously. Still better than switching off.

theosanderson:
CI should be self-contained. If it says there's something wrong that should indicate there's something wrong with the underlying code (or with some infrastructure, or something). Loads of PRs in this repo will be like fixing a typo from `gnomic` to `genomic`. They don't need previews. But as I understand it atm, without previews this will yield a failed check. It's really nice to be able to trust that most of the time a failed check means something.

theosanderson:
But on reflection maybe I'm misunderstanding and it only checks if there actually is a preview up?

theosanderson:
Ah yeah, no here's what I was referring to: https://github.com/pathoplexus/pathoplexus/actions/runs/10893951473/job/30229904944

corneliusroemer:
In that case it saya the preview was down - and there was a label added. If we have a preview label it's good to check the preview actually works no?

theosanderson:
👍  OK thanks for explaining - I guess the preview hadn't come up due to resource constraints. To me that's an indication that ultimately this would be nice to do fully in GitHub Actions at some point - but I agree that I wasn't understanding the situation and that this seems fine for now (if we sort out the flaking on links themselves)

corneliusroemer:
I think it's good to have a check that previews that we request are up - as long as we know that link check also indicates that we might be able to work with that?

I agree it's nice if we can do link checks without preview but for now that seems not worth the work?

theosanderson:
It only seems worth the effort once we are doing E2E tests anyway and thus isn't substantial extra effort.

PR #154, opened 2024-09-18
fix: don't ingest continent as geoLocAdmin2

corneliusroemer:
This should be released to production together with the author order fix so we only do one version bump


anna-parker (approved):

corneliusroemer:
Merging this - it doesn't have any impact on production by itself as the values there are decoupled

PR #166, opened 2024-09-24
Try to fix broken link checker

### LINKED ISSUE #165: [BUG] Broken link checker wrongly reports errors for links with anchor
[Here](https://github.com/pathoplexus/pathoplexus/actions/runs/11022387713/job/30611831355), the broken link checker reports the following two links as not found although they work:

- https://backend.pathoplexus.org/swagger-ui/index.html#/submission-controller/submit
- https://backend.pathoplexus.org/swagger-ui/index.html#/submission-controller

image

theosanderson:
Resolves #165

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -66,4 +66,4 @@ jobs:
         with:
           url: ${{ steps.set-url.outputs.url }}
           # see https://github.com/raviqqe/muffet for details
-          cmd_params: '--buffer-size=8192 --max-connections=10 --color=always --accepted-status-codes="200..300,403,429" --rate-limit=20 --timeout=20'
+          cmd_params: '-exclude=(#/) --buffer-size=8192 --max-connections=10 --color=always --accepted-status-codes="200..300,403,429" --rate-limit=20 --timeout=20'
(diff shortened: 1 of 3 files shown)


.github/workflows/broken-link-check.yml
corneliusroemer, line 1: You should just use `--ignore-fragments`

It's a bug in swagger that it doesn't show the IDs in the static HTML it generates but we can work around it like this

.github/workflows/broken-link-check.yml
```diff
@@ -66,4 +66,4 @@ jobs:
         with:
           url: ${{ steps.set-url.outputs.url }}
           # see https://github.com/raviqqe/muffet for details
-          cmd_params: '--buffer-size=8192 --max-connections=10 --color=always --accepted-status-codes="200..300,403,429" --rate-limit=20 --timeout=20'
+          cmd_params: '-exclude="#/" --buffer-size=8192 --max-connections=10 --color=always --accepted-status-codes="200..300,403,429" --rate-limit=20 --timeout=20'
```
corneliusroemer, line 69:
```suggestion
          cmd_params: '--buffer-size=8192 --max-connections=10 --color=always --accepted-status-codes="200..300,403,429" --rate-limit=20 --timeout=20 --ignore-fragments'
```

.github/workflows/broken-link-check.yml
corneliusroemer: You used a single hyphen for exclude that's one reason why the exclude didn't work

corneliusroemer:
Made it work

PR #167, opened 2024-09-25
Add clone scripts to the repo for versioning

corneliusroemer:
To clone from prod to staging, run `./clone-prod-to-staging`

ssh into bastion: 
```
ssh bastion
cd ./pathoplexus/scripts/db-clone/
./clone-prod-to-staging
```

### DIFF
--- a/scripts/db-clone/README.md
+++ b/scripts/db-clone/README.md
@@ -16,6 +16,7 @@ Passwords for the following users need to be set in the environment variables:
 - `staging_loculus_user`
 - `prod_keycloak_user`
 - `staging_keycloak_user`
+- `postgres`
 
 Run the script as follows:
 
--- a/scripts/db-clone/README.md
+++ b/scripts/db-clone/README.md
@@ -21,7 +21,7 @@ Passwords for the following users need to be set in the environment variables:
 Run the script as follows:
 
 ```sh
-./scripts/db-clone/clone-prod-to-staging.sh
+./clone-prod-to-staging.sh
 ```
 
 It can happen that part of the clone fails. You can check the logs if things don't work as expected.
--- a/scripts/db-clone/README.md
+++ b/scripts/db-clone/README.md
@@ -1,0 +1,26 @@
+# Script to clone production database to staging database
+
+This script is used to make the staging database tables a clone of the production database tables. A few modifications are made to the dumps before loading so that the staging database works (e.g. users are renamed, ownership changed).
+
+The script roughly follows these steps:
+
+1. Dump the production database tables.
+2. Modify the dump to rename users and change ownership.
+3. Load the modified dump into the staging database tables.
+
+The script needs to be run on a server with access to the database (e.g. a bastion host on AWS).
+
+Passwords for the following users need to be set in the environment variables:
+
+- `prod_loculus_user`
+- `staging_loculus_user`
+- `prod_keycloak_user`
+- `staging_keycloak_user`
+
+Run the script as follows:
+
+```sh
+./scripts/db-clone/clone-prod-to-staging.sh
+```
+
+It can happen that part of the clone fails. You can check the logs if things don't work as expected.
--- a/scripts/db-clone/clone-prod-to-staging.sh
+++ b/scripts/db-clone/clone-prod-to-staging.sh
@@ -27,12 +27,13 @@ perform_sed_replacements() {
     local file="$1"
     echo "Performing sed replacements on $file..."
     # Abort the script if a line with `@` contains the word "prod" to prevent altering sensitive data
-    if grep '@' "$file" | grep -q 'prod'; then
+    if grep '@' "$file" | grep -q 'prod_'; then
         echo "Error: Found 'prod' in line with '@' in $file. Aborting to prevent changing sensitive data."
         exit 1
     fi
 
-    sed -i 's/prod/staging/g' "$file"
+    sed -i 's/prod_loculus_user/staging_loculus_user/g' "$file"
+    sed -i 's/prod_keycloak_user/staging_keycloak_user/g' "$file"
     sed -i 's#//pathoplexus.org#//staging.pathoplexus.org#g' "$file"
     sed -i 's#authentication.pathoplexus.org#authentication-staging.pathoplexus.org#g' "$file"
     check_command "Failed to perform sed replacements on $file"
--- a/scripts/db-clone/clone-prod-to-staging.sh
+++ b/scripts/db-clone/clone-prod-to-staging.sh
@@ -27,11 +27,9 @@ perform_sed_replacements() {
     local file="$1"
     echo "Performing sed replacements on $file..."
     # Abort the script if a line with `@` contains the word "prod" to prevent altering sensitive data
-    if grep -q '@' "$file"; then
-        if grep -q 'prod' "$file"; then
-            echo "Error: Found 'prod' in $file. Aborting to prevent changing sensitive data."
-            exit 1
-        fi
+    if grep '@' "$file" | grep -q 'prod'; then
+        echo "Error: Found 'prod' in $file. Aborting to prevent changing sensitive data."
+        exit 1
     fi
 
     sed -i 's/prod/staging/g' "$file"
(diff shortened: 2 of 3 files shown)


scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Could screw up columns and values that contain `prod` etc
+# For now not an issue but might eventually want to be more surgical
+perform_sed_replacements() {
+    local file="$1"
+    echo "Performing sed replacements on $file..."
+    sed -i 's/prod/staging/g' "$file"
+    sed -i 's#//pathoplexus.org#//staging.pathoplexus.org#g' "$file"
+    sed -i 's#authentication.pathoplexus.org#authentication-staging.pathoplexus.org#g' "$file"
+    check_command "Failed to perform sed replacements on $file"
```
theosanderson, line 34: If I understand correctly this will only check the last `sed` command? I kinda think it can be safer to use `set -e` at the top of the script

scripts/db-clone/clone.sh
```diff
@@ -0,0 +1,154 @@
...
+    local db_name="$1"
+    local dump_file="$2"
+    echo "Dumping $db_name database to $dump_file..."
+    pg_dump "$db_name" > "$dump_file"
+    check_command "Failed to dump $db_name database"
+    echo "Database dump completed successfully!"
+    echo "Dump file: $dump_file"
+}
+
+prepare_database() {
```
theosanderson, line 36: I'd find a name like `drain_database_connections` clearer

scripts/db-clone/clone.sh
```diff
@@ -0,0 +1,154 @@
...
+    local db_name="$1"
+    local new_owner="$2"
+    echo "Changing ownership to $new_owner"
+    psql -d "$db_name" <<EOF
+    DO \$\$
+    DECLARE
+        r RECORD;
+    BEGIN
+        -- Change ownership of tables
+        FOR r IN (SELECT schemaname, tablename FROM pg_tables WHERE schemaname = 'public')
```
theosanderson, line 82: Couple of things to think about:

1. If in the future we add anything else like Materialized Views we'll need to remember to add them here. In an ideal world we might use another approach that didn't require us to iterate over the objects if that was possible. E.g. if they were created by their eventual owner? But this may be complicated
2. I think @anna-parker already uses a second schema in addition to `public` ?

fhennig:
I think it's nice to start shell scripts with 

```
#!/usr/bin/env bash
set -eu
```

- using `env` is more portable (but there are also cons to using it, I read online)
- `-e` exits on errors
- `-u` treats unset vars as errors

Just preference though, feel free to ignore!

scripts/db-clone/clone.sh
anna-parker: yes - currently ena-submission is disable so that means the schema has not been created yet on prod/staging but as soon as it has been we will have an additional schema called `ena-submission`

scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -0,0 +1,56 @@
...
+# Set variables
+CHILD_SCRIPT="./clone.sh"
+PROD_KC_DUMP="production_keycloak_dump.sql"
+PROD_LOC_DUMP="production_loculus_dump.sql"
+PROD_KC_DB="pathoplexus_prod_keycloak"
+PROD_LOC_DB="pathoplexus_prod_loculus"
+STAGING_KC_DB="pathoplexus_staging_keycloak"
+STAGING_LOC_DB="pathoplexus_staging_loculus"
+STAGING_KC_USER="staging_keycloak_user"
+STAGING_LOC_USER="staging_loculus_user"
```
chaoran-chen, line 15: It doesn't require an username and password for the production database? What are the assumptions / required environment?

scripts/db-clone/clone.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+
+set -x
+
+
+export PGHOST=pathoplexus3.cluster-che6ccwa8oie.eu-central-1.rds.amazonaws.com
+export PGUSER=postgres
```
chaoran-chen, line 7: For reading the production database, could we create a read-only user? postgres is really powerful and we should avoid using it as much as possible.

scripts/db-clone/clone-prod-to-staging.sh
chaoran-chen, line 1: I think that it would be very useful to have a documentation (e.g., a README file in the folder) that explains the purpose of this script. Further, either here or in a different repository, it would be great to document from which server/place this should be executed.

scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -0,0 +1,56 @@
...
+
+check_command() {
+    if [ $? -ne 0 ]; then
+        echo "Error: $1"
+        exit 1
+    fi
+}
+
+# Note: Could screw up columns and values that contain `prod` etc
+# For now not an issue but might eventually want to be more surgical
```
chaoran-chen, line 25: Is this not critical and can it not potentially be used for an attack? If someone, for example, has "prod" in their email (let's say `prodigy@gmail.com`) and this get's replaced to `stagingidy@gmail.com`, someone can potentially create the `stagingidy@gmail.com` address to reset the password and get access to the account. In combination with https://github.com/pathoplexus/private_issues/issues/7#issuecomment-2371978447, this can potentially even allow access to the account on the production server.

chaoran-chen (dismissed):

scripts/db-clone/clone-prod-to-staging.sh
chaoran-chen: Could we instead first load the new database and then use SQL for a few targeted replacements?

corneliusroemer:
The latest few commits make sure we don't replace "prod" in lines that contain email addresses (contain `@`) - feel free to check @chaoran-chen

scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -0,0 +1,62 @@
...
+perform_sed_replacements() {
+    local file="$1"
+    echo "Performing sed replacements on $file..."
+    # Abort the script if a line with `@` contains the word "prod" to prevent altering sensitive data
+    if grep '@' "$file" | grep -q 'prod'; then
+        echo "Error: Found 'prod' in line with '@' in $file. Aborting to prevent changing sensitive data."
+        exit 1
+    fi
+
+    sed -i 's/prod/staging/g' "$file"
```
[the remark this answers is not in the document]
anna-parker: I think the only place that we would need to perform this change is in the file names, but maybe we could replace this line with `sed -n '/prod/ s/prod/staging/gp' "$file"` so that we can see in the logs where this replacement occurs

scripts/db-clone/clone.sh
theosanderson: @anna-parker is it still disabled? Do we need to change anything here do you think or is this good enough as is?

scripts/db-clone/clone-prod-to-staging.sh
corneliusroemer: We can get rid of the check command

scripts/db-clone/clone-prod-to-staging.sh
corneliusroemer: Done

scripts/db-clone/README.md
```diff
@@ -0,0 +1,26 @@
...
+3. Load the modified dump into the staging database tables.
+
+The script needs to be run on a server with access to the database (e.g. a bastion host on AWS).
+
+Passwords for the following users need to be set in the environment variables:
+
+- `prod_loculus_user`
+- `staging_loculus_user`
+- `prod_keycloak_user`
+- `staging_keycloak_user`
```
corneliusroemer, line 18:
```suggestion
- `staging_keycloak_user`
- `postgres`
```

scripts/db-clone/README.md
```diff
@@ -0,0 +1,27 @@
...
+- `postgres`
+
+Run the script as follows:
+
+```sh
+./scripts/db-clone/clone-prod-to-staging.sh
```
corneliusroemer, line 24:
```suggestion
./clone-prod-to-staging.sh
```

scripts/db-clone/clone.sh
theosanderson: Actually it looks like this function is unused? OK to delete it? Is it the same for `change_ownership` and the `restore_access` commands?

PR #179, opened 2024-09-26
feat: reserve more resources for prod

corneliusroemer:
We weren't reserving much memory for many prod pods, this fixes it.

Also, I don't think we need to reserve so much cpu for lapis/silo, these don't usually require much cpu and we're more cpu constrained at the moment in terms of requests than memory.

### DIFF
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -101,7 +101,7 @@ resources:
   lapis:
     requests:
       memory: "220Mi"
-      cpu: "500m"
+      cpu: "1000m"
     limits:
       memory: "5Gi"
   silo-preprocessing:
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -88,14 +88,14 @@ resources:
       memory: "10Gi"
   keycloak:
     requests:
-      memory: "3Gi"
+      memory: "1500Mi"
       cpu: "1000m"
     limits:
       memory: "3Gi"
   silo:
     requests:
       memory: "500Mi"
-      cpu: "500m"
+      cpu: "1000m"
     limits:
       memory: "10Gi"
   lapis:
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -69,10 +69,10 @@ podPriorityClassName: production
 resources:
   website:
     requests:
-      memory: "200Mi"
+      memory: "1Gi"
       cpu: "1000m"
     limits:
-      memory: "1Gi"
+      memory: "5Gi"
   ena-submission:
     requests:
       memory: "80Mi"
@@ -88,20 +88,20 @@ resources:
       memory: "10Gi"
   keycloak:
     requests:
-      memory: "500Mi"
+      memory: "3Gi"
       cpu: "1000m"
     limits:
       memory: "3Gi"
   silo:
     requests:
-      memory: "100Mi"
-      cpu: "1000m"
+      memory: "500Mi"
+      cpu: "500m"
     limits:
       memory: "10Gi"
   lapis:
     requests:
       memory: "220Mi"
-      cpu: "1000m"
+      cpu: "500m"
     limits:
       memory: "5Gi"
   silo-preprocessing:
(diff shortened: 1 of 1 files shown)


loculus_values/environment_specific_values/production.yaml
```diff
@@ -69,10 +69,10 @@ podPriorityClassName: production
 resources:
   website:
     requests:
-      memory: "200Mi"
+      memory: "1Gi"
       cpu: "1000m"
     limits:
-      memory: "1Gi"
+      memory: "5Gi"
```
theosanderson, line 75: This may be necessary ATM due to non-ideal coding. I think we should aim to limit at less than this in the future (going more for having many replicas). [You can resolve this comment immediately, not requesting changes]

loculus_values/environment_specific_values/production.yaml
```diff
@@ -88,20 +88,20 @@ resources:
       memory: "10Gi"
   keycloak:
     requests:
-      memory: "500Mi"
+      memory: "3Gi"
```
theosanderson, line 91: This seems a bit too much per https://www.keycloak.org/high-availability/concepts-memory-and-cpu-sizing#:~:text=Recommendations%3A,connects%20to%20only%20one%20client. but I may be misunderstanding that?

loculus_values/environment_specific_values/production.yaml
```diff
@@ -88,20 +88,20 @@ resources:
...
       cpu: "1000m"
     limits:
       memory: "3Gi"
   silo:
     requests:
-      memory: "100Mi"
-      cpu: "1000m"
+      memory: "500Mi"
+      cpu: "500m"
```
theosanderson, line 98: Idk, I assume that these don't require much CPU except when they do when they're being heavily used (e.g. once someone starts using them as an API). To me giving 1cpu each in prod doesn't seem excessive.

corneliusroemer: We'd be having 8 CPUs idle just for LAPIS/SILO, just wondering whether you're aware of that. If you are fine with me.

theosanderson: > We'd be having 8 CPUs idle just for LAPIS/SILO, just wondering whether you're aware of that.

To be pedantic, that's not true in my understanding. We would be saying that production LAPIS/SILO would be guaranteed 8 CPUs if it needed them. When this CPU isn't being used by LAPIS/SILO it can be used by other stuff, e.g. preview instances (i.e. the difference between the preview instance's `requests` and their `limit`).

loculus_values/environment_specific_values/production.yaml
corneliusroemer: We have plenty of RAM, doesn't harm to reserve it - but yeah we could reduce if you like. We're not RAM constrained right now though.

loculus_values/environment_specific_values/production.yaml
corneliusroemer: Ah yeah you're right, sorry. However it does mean we do require a minimum number of CPUs - but that's fine.

loculus_values/environment_specific_values/production.yaml
corneliusroemer: Indeed, only necessary due to lack of pagination - but given we require that much RAM sometimes we should also grant it.

loculus_values/environment_specific_values/production.yaml
corneliusroemer, line 91:
```suggestion
      memory: "1500Mi"
```

loculus_values/environment_specific_values/production.yaml
```diff
@@ -88,20 +88,20 @@ resources:
...
-      memory: "500Mi"
+      memory: "3Gi"
       cpu: "1000m"
     limits:
       memory: "3Gi"
   silo:
     requests:
-      memory: "100Mi"
-      cpu: "1000m"
+      memory: "500Mi"
```
corneliusroemer, line 97:
```suggestion
      memory: "1000Mi"
```

loculus_values/environment_specific_values/production.yaml
corneliusroemer, line 98:
```suggestion
      cpu: "1000m"
```

loculus_values/environment_specific_values/production.yaml
corneliusroemer: I'll reduce it to 1.5GB

loculus_values/environment_specific_values/production.yaml
```diff
@@ -88,20 +88,20 @@ resources:
...
-      memory: "100Mi"
+      memory: "500Mi"
       cpu: "1000m"
     limits:
       memory: "10Gi"
   lapis:
     requests:
       memory: "220Mi"
-      cpu: "1000m"
+      cpu: "500m"
```
corneliusroemer, line 104:
```suggestion
      cpu: "1000m"
```

theosanderson (approved):
Thanks!

PR #180, opened 2024-09-26
Adding Meera to SAB page

emmahodcroft:
~~[DO NOT MERGE WITHOUT EMMA'S EXPLICIT OK]~~

In preparation for adding Meera to SAB page, here's a PR doing this adding, though it only uses a silouette picture as there are none of her available. She'll send one when she has one. I've also send a screenshot to her so she can confirm it looks ok.

However, Meera still needs to go through the final steps of recommendation & EB vote, so this should not be merged anytime soon.


emmahodcroft:
Meera has approved the UKHSA works as an affiliation. She still doesn't have a headshot, but will update us when she does. We can plan to move forward with the silhouette for the time being, after approval at the EB next week :)

emmahodcroft:
Meera's been approved so this could be merged now. I'll turn off preview.

Branch is now out of date, even though this is a minor change that won't impact anything else. if someone could 'make updated' that would be amazing.

fhennig (approved):
LGTM

Issue #152, opened 2024-09-17
[BUG] CI link-checker is flaky


anna-parker:
**Describe the bug**
Our link checker test keeps failing with timeouts: 
<img width="719" alt="image" src="https://github.com/user-attachments/assets/7bbba16a-2199-4e46-8d2f-b7e4726d397a">

Lets see if we can increase timeout in the config somewhere

corneliusroemer:
Thanks for investigating!

Let's keep track of which URLs time out - if it's external ones we can simply filter them out using regex.

Please add any that time out to this issue and I'll fix it.

Also, links to the action/step that failed would be great to include in the issue 🙏

corneliusroemer:
I think I've mostly deflaked it now - reopen if the issue is back.

PR #187, opened 2024-09-29
Increase timeout length for muffet

theosanderson:
Resolves(?) #152

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -44,7 +44,7 @@ jobs:
         if: github.event_name == 'push' || github.event_name == 'pull_request'
         run: |
           echo "Waiting for 180 seconds to allow preview to be up..."
-          sleep 180
+          # sleep 180
   broken-external-link-checker:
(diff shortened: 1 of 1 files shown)


theosanderson:
I don't really see what is wrong with my original version here? I think trying to test for Loculus stuff that takes too long, if that's your goal, doesn't make sense on preview servers which can be throttled

theosanderson:
(but I also don't have a problem with trying it, now you've got it passing)

theosanderson:
The downside is just the code duplication

.github/workflows/broken-link-check.yml
```diff
@@ -44,8 +44,8 @@ jobs:
         if: github.event_name == 'push' || github.event_name == 'pull_request'
         run: |
           echo "Waiting for 180 seconds to allow preview to be up..."
-          sleep 180
-  broken-link-checker:
+          # sleep 180
```
corneliusroemer, line 47:
```suggestion
          sleep 180
```

PR #195, opened 2024-09-30
Add custom github link for pathoplexus

theosanderson:
Relates to https://github.com/loculus-project/loculus/pull/2797

~(preview not needed IMO, this is currently being merged in advance of the loculus bump, which is fine)~

https://preview-custom-github-link.pathoplexus.org/


corneliusroemer (approved):
Something off with preview but the GitHub button works 😀

PR #196, opened 2024-09-30
Increase muffet timeout again

theosanderson:
This goes back to my original commit from https://github.com/pathoplexus/pathoplexus/pull/187 as, for whatever reason, what's on main isn't working. Also it seems to include some internal links in external? https://github.com/pathoplexus/pathoplexus/actions/runs/11111325516/job/30871119683

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           echo "Waiting for 180 seconds to allow preview to be up..."
           sleep 180
-  broken-external-link-checker:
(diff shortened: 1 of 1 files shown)


theosanderson:
Closing for now - the issue with main is insufficient LAPIS memory

corneliusroemer:
It's not possible to only test external links, external is a superset of
internal.

theosanderson:
hmm - then I'd really not split them out (and if we did, name them appropriately)

corneliusroemer:
the splitting is so that we have more and less flaky actions. happy to rename: internal-only, external-also?

PR #200, opened 2024-10-02
Add gcaAccession

anna-parker:
The ena-submission pipeline is failing to upload external metadata because it is uploading a gcaAccession field which is currently not in the config. 

I hope that this will not cause temporary downtime as this field is not submitted by users but we should test on staging first especially as addng latitude and longitude was leading to issues in LAPIS-SILO.

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -486,6 +486,14 @@ defaultOrganismConfig: &defaultOrganismConfig
         header: "INSDC"
         noInput: true
         oneHeader: true
+      - name: gcaAccession
+        displayName: GCA accession
+        customDisplay:
+          type: link
+          url: "https://www.ncbi.nlm.nih.gov/datasets/genome/__value__"
+        header: "INSDC"
+        noInput: true
+        oneHeader: true
       - name: cultureId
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
Nice, shall we also make that change over at loculus, or is it already done there? Or does it not make sense there?

corneliusroemer:
Let's test on staging before merging to main to reduce chance this ends up broken in prod accidentally?

theosanderson:
My instinct is to be surprised if it's not a breaking change - but we'll find out!

corneliusroemer:
> My instinct is to be surprised if it's not a breaking change - but we'll find out!

What do you think would break @theosanderson? This is post preprocessing IIUC

corneliusroemer:
@anna-parker I will add a github action that forbids underscores in branch names, that's why the preview fails 😝

theosanderson:
I probably just don't understand. If you guys are reasonably confident then probably makes sense to see what happens on staging.

theosanderson:
I think a GHA to forbid underscores is too late - any such efforts would probably be better invested in sanitizing the name in the applicationset or something

PR #201, opened 2024-10-02
Add gcaAccession

anna-parker:
Update: https://github.com/loculus-project/loculus/pull/3232 fixed the issue that previously blocked up from adding new fields without reprocessing all metadata

Note we will still need to update the external metadata pipeline to upload the gcaAccession to PPX - but we can do that in a later PR

https://preview-add-gca-accession.ppx.bio

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -679,6 +679,14 @@ defaultOrganismConfig: &defaultOrganismConfig
         definition: "Accession of corresponding public BioSample of the raw reads in the INSDC"
         example: "SAMN12345678"
         oneHeader: true
+      - name: gcaAccession
+        displayName: GCA accession
+        customDisplay:
+          type: link
+          url: "https://www.ncbi.nlm.nih.gov/datasets/genome/__value__"
+        header: "INSDC"
+        noInput: true
+        oneHeader: true
       - name: cultureId
         displayName: Culture ID
         header: Sample details


anna-parker:
Actually no it is probably exactly the same as adding latitude/longitude. Even noInput fields are passed to preprocessing. So I assume the backend needs prepro to add the {gcaAccession: null} to the metadata that is stored in the backend. 

So we should increase the prepro pipeline by 1 version.

anna-parker:
> What happens if both prepro and external metadata supply a field?

1. Currently if a field exists in the sequence_entries and in external metadata it will be overwritten by the value in external metadata 

> Can we make a field external-metadata-only? Right now, IIUC, prepro config contains all metadata fields from values yaml.
Where does external metadata get its fields to submit from? Hard coded? Or configurable? Probably hard coded.

3. That would be nice, currently the config just creates the external metadata list from all fields that have the header "INSDC".

corneliusroemer:
> Even noInput fields are passed to preprocessing. So I assume the backend needs prepro to add the {gcaAccession: null} to the metadata that is stored in the backend. Yes see my comment here: https://github.com/pathoplexus/pathoplexus/pull/139#issuecomment-2357874144 - the backend does not modify the data it sends to LAPIS/SILO.

I'd like to understand the reasoning that leads you to that conclusion so I can see where I'm wrong. Which of my premises is wrong? Do you think backend will error? If so why and where?

anna-parker:
> Even noInput fields are passed to preprocessing. So I assume the backend needs prepro to add the {gcaAccession: null} to the metadata that is stored in the backend. Yes see my comment here: https://github.com/pathoplexus/pathoplexus/pull/139#issuecomment-2357874144 - the backend does not modify the data it sends to LAPIS/SILO.

SILO/LAPIS will receive a config update, this config will require metadata to have a field called gcaAccession. The backend will send SILO/LAPIS the metadata it has stored in the backend. This will be old data, it will not include the gcaAccession. The new SILO/LAPIS pod will not be able to start. The backend will now accept external metadata updates with the gcaAccession field but unless an entry is updated it will not be added. 

The prepro config will receive gcaAccession as a field and any new processed data will include this field (by default set to null). Updating the prepro version will cause all data in the database to be reprocessed and all old fields will receive the gcaAccession: null field. After this succeeds the new LAPIS/SILO pod will start.

corneliusroemer:
Thanks @anna-parker - so IIUC, my wrong assumption above is that:

> A2. External metadata will submit the field to backend for all sequences next time it runs

Is that not correct? I thought we run the external metadata submission frequently to make use of newly available data. Otherwise, how can we put new external metadata into the backend? Is there no mechanism? Relying on prepro to fix something that's coming from external metadata seems brittle.

anna-parker:
So the external_metadata submission endpoint just adds metadata to a table in the backend. Metadata fields will only be accepted if they are in the external metadata list. (which of course will change when the config changes - but the data that is already in the table is persistent)

Tdlr the sequence_entries_view will join metadata from the external_metadata_view table and the sequence_entries_preprocessed_data, if a field is in both tables and not null we will choose the entry from external_metadata_view. So this means for any data, including external metadata entries submitted prior to the config update will fulfill the requirements of the previous config.

anna-parker:
This would be quite a big change - currently the external metadata updater only sends update requests to the backend when it has an update.

I think your suggested change would mean the external metadata updater will always have to send empty update requests for every accession, version in the the database with all external metadata fields in the config - which seems like a lot of load with little benefit.

corneliusroemer:
Nice, so we can just add the field, and then have ena-submission submit is as external metadata. What are the required config changes? Adding a field is simple, we'd probably want it to be marked as a special "warning" field in case someone who isn't ingest submits it directly.

corneliusroemer (approved):

PR #206, opened 2024-10-03
Add auth for staging

theosanderson:
Adds `htpasswd` for basic auth


loculus_values/values.yaml
```diff
@@ -25,6 +25,11 @@ secrets:
...
+  basicauth:
+    type: sealedsecret
+    clusterWide: true
+    encryptedData:
+      users: AgBfZyTygLwse/CBQx+qy2tY+pfcFulezoyqFOofNu/ZkKfIokObxiUYPVKBHwn2QavQgU7TK//NNLylADNUVwi5rWbRP0akOJYLb48h8GaRF3dyRp6t4OhB3N0obuaKmVyYW0C31AzabdLfB7//E4Xzg/TWbaz+p7vgYD9SPNb4/Qk1svDoLNKOi10Ghdi8zR8YB5Sje5p+ah0to21K3YfHdg68y13AirIwcQqo1PmpO9jcbwSy9y00A71vffnmYMXIqBLppriPnFs/4p/OtBKz73eRE7GcRkNt1lfxmiv5uOB24tBDHbTENpzGeDqdWLCPu7aa5Vwt6JQFz4pXtWVVPpsxRYtRGx96uadoLG3Za5fN66TADUEatxvNDbiWv+421qwdgogFYvEa7LWISkwKI2Hm6/wpyOJysaLKzPl8159l/tavXQyphMamoaxiqDNPyDYJ/cbMV4TUBMd80zfQMep2PPcnT9oxJcMFKeFJSEkEgJ2LCE58Wrp/N01FPvmp5rQDrTCAvqJE4xL3TepXoGEBzgYN2coF5ChL4r1wdCd/3l6PdV6mLHanu26xJ7XEF8/fKOX1pCtNcIBQGvr2pX3NTii54Sp0b6YvDpgCdi2S0Q1mjqwG4ZZNkvrj1rH7KigxPtl+1k/4Pz8a2X/JMq5rubXCa+QzsAxZZtlK2hRa1CrA0/f3nZdX7jHIpSS7H5H48QF7CpWghq/SwE9jgk74z/xwweSrpp33CPTTFKmgE5PGc4iD
```
corneliusroemer, line 32: Can you explain what is inside the encrypted stuff (except for the literal secret of course) - this looks a bit cryptic to me!

corneliusroemer (approved):

loculus_values/values.yaml
theosanderson: Made by https://httpd.apache.org/docs/2.4/programs/htpasswd.html - not 100% sure if it will work

corneliusroemer: Ok I get it now. Basic auth middleware expects a spec like this:

```
apiVersion: traefik.containo.us/v1alpha1
kind: Middleware
metadata:
  name: basic-auth
spec:
  basicAuth:
    users: admin:$$apr1$$xxxxxxxx$$xx.xxxxxx/xxxxx/xxx
```
where `admin...` is an apache htpasswd string that encode username and password.

Now rather than literally encoding it like this, we use a secret to sub in this keyval pair `users: admin:$$apr1$$xxxxxxxx$$xx.xxxxxx/xxxxx/xxx`.

We could have done this using a raw secret:

```
  basicauth:
    type: raw
    data:
      users: admin:$$apr1$$xxxxxxxx$$xx.xxxxxx/xxxxx/xxx
```

but of course then it's not secret.

What we did for testing was more convenient but essentially the same:

```
  basicauth:
    type: rawhtpasswd
    data:
      username: loculus
      password: widetailpotato
```

which gets transformed into exactly the same string as the equivalent raw one, just more convenient as one doesn't need to run through `htpasswd`.

However we can't do this with a sealed secret of course.

So now we need to seal the original htpasswd encoded string. Which is what you're doing here.

PR #207, opened 2024-10-03
Remove staging from muffet link check

theosanderson:
Staging will be broken by auth

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -37,7 +37,7 @@ jobs:
               echo "run_check=false" >> $GITHUB_OUTPUT
             fi
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/broken-link-check.yml
theosanderson, line 1: Actually shouldn't this CI be on loculus-deployments?

theosanderson: (in part)

corneliusroemer: If we want we can trigger it there upon bump - no need to be located there

corneliusroemer (approved):

corneliusroemer (approved):

PR #222, opened 2024-10-07
ENA deposition config updates, required by loculus changes

anna-parker:
Preview: https://preview-update-ena-configs.pathoplexus.org/

This includes the config updates from https://github.com/loculus-project/loculus/pull/2960

We can test this works first by uploading the third sequence to the ena-submission test folder: https://github.com/pathoplexus/ena-submission/pull/8


theosanderson:
The approach I took was to merge the config changes into the bumping PR. I think that can be nice as it means that at every commit to main the deployment should be properly working. In due course we'll have E2E tests for PP and so that will be somewhat necessary.

corneliusroemer (approved):

PR #239, opened 2024-10-11
Add new required authors format to docs

anna-parker:
Needs to be merged with https://github.com/loculus-project/loculus/pull/2986


corneliusroemer:
I think you merged this already separately so we can close this I think, right @anna-parker? :)

PR #241, opened 2024-10-14
docs: document group edit feature

fhennig:
This PR also adds a "Caution box", an _aside_ or _admonition_. 

I have also changed the ordering to be spread out a bunch more (order numbers contain gaps) to make it easier to add documents in the future.

The styling of some links is also changed.

https://preview-update-docs-group-edit.pathoplexus.org/docs/how-to/edit-group

### DIFF
--- a/monorepo/website/src/components/asides/InfoBox.astro
+++ b/monorepo/website/src/components/asides/InfoBox.astro
@@ -2,8 +2,8 @@
 import MaterialSymbolsInfoOutline from '~icons/material-symbols/info-outline';
 ---
 
-<div class='bg-primary-500 !text-white p-3 my-4 rounded-lg max-w-md mx-auto box text-sm'>
-    <span style={{ fontSize: '1.0rem' }} class='!text-white'>
+<div class='bg-primary-300 p-3 my-4 rounded-lg mx-auto box text-sm'>
+    <span style={{ fontSize: '1.0rem' }}>
         <MaterialSymbolsInfoOutline />
     </span>
     <slot />
--- a/monorepo/website/src/components/asides/InfoBox.astro
+++ b/monorepo/website/src/components/asides/InfoBox.astro
@@ -13,8 +13,11 @@ import MaterialSymbolsInfoOutline from '~icons/material-symbols/info-outline';
 </style>
 
 <div class='bg-primary-200 p-3 my-4 rounded-lg mx-auto box text-sm info-box-link'>
-    <span style={{ fontSize: '1.0rem' }}>
-        <MaterialSymbolsInfoOutline />
+    <span class='text-base inline-flex items-center'>
+        <span class='inline-block mr-2'>
+            <MaterialSymbolsInfoOutline />
+        </span>
+        Note
     </span>
     <slot />
 </div>
--- a/monorepo/website/src/components/asides/InfoBox.astro
+++ b/monorepo/website/src/components/asides/InfoBox.astro
@@ -2,7 +2,17 @@
 import MaterialSymbolsInfoOutline from '~icons/material-symbols/info-outline';
 ---
 
-<div class='bg-primary-300 p-3 my-4 rounded-lg mx-auto box text-sm'>
+<style is:global>
+    .info-box-link a {
+        @apply text-black;
+    }
+
+    .info-box-link a:hover {
+        @apply text-primary-800;
+    }
+</style>
+
+<div class='bg-primary-200 p-3 my-4 rounded-lg mx-auto box text-sm info-box-link'>
     <span style={{ fontSize: '1.0rem' }}>
         <MaterialSymbolsInfoOutline />
     </span>
--- a/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
+++ b/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
@@ -10,7 +10,8 @@ When making this request, please be sure to supply your username exactly as it a
 
 You can then ask them to follow the below steps to add you to their group.
 
-## Adding a user to your group
+## Adding a user to a group you are a member of
+
 
 1. Log into your account and navigate to your account page ('My account' in the top-right of the website)
 2. Within your groups, select the name of the group that you'd like to add a user to
--- a/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
+++ b/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
@@ -1,20 +1,23 @@
 ---
 layout: ../../../layouts/DocLayout.astro
 title: "Adding or removing a user group"
-order: 5
+order: 21
 ---
 
-Users cannot add themselves to groups; they must ask someone within a group to add them. When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
+Users cannot add themselves to groups; they must ask someone within a group to add them.
+When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
 
 You can then ask them to follow the below steps to add you to their group.
 
-To add a user to your group:
+## Adding a user to your group
+
 1. Log into your account and navigate to your account page ('My account' in the top-right of the website)
 2. Within your groups, select the name of the group that you'd like to add a user to
 3. Using the 'user' interface at the bottom of the page, type in the username of the person you want to add
 4. The user is now part of your group
 
-To remove a user:
+## Removing a user from your group
+
 1. Log into your account and navigate to your account page ('My account' in the top-right of the website)
 2. Within your groups, select the name of the group that you'd like to remove a user from
 3. Using the 'user' interface at the bottom of the page, click the red button next to the user you wish to remove
--- a/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
+++ b/monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
@@ -4,7 +4,8 @@ title: "Adding or removing a user group"
 order: 21
 ---
 
-Users cannot add themselves to groups; they must ask someone within a group to add them.
+Users cannot add themselves to groups; they must ask someone within a group to add them. (Users are automatically members of groups they create)
+
 When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
 
 You can then ask them to follow the below steps to add you to their group.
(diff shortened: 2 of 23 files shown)


monorepo/website/src/pages/docs/how-to/edit-group.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+Any member of a group can edit group information, which means that you can edit group information for all the groups that you are a member of.
+
+1. Log into your account and enter your account page by clicking 'My account' in the top-right corner of the webpage
+2. Under 'My groups' click the group that you want to edit
+3. In the group detail view, click the 'Edit group' button in the top right corner
+4. Edit any property of the group you want to change and click 'Update group' to save your changes
+
+<CautionBox>
+If you change the name of a group the center name in corresponding ENA submissions will also change.
```
theosanderson, line 16: (cool PR) This implies to me that the change will propagate to existing submissions which I would guess it won't

fhennig: Uhm, I think I'm not too familiar with how this ENA stuff works, I'd ask @corneliusroemer to weigh in? We have the same line of text in the loculus docs, and there Cornelius was part of coming up with the text.

monorepo/website/src/components/asides/InfoBox.astro
```diff
@@ -2,8 +2,8 @@
 import MaterialSymbolsInfoOutline from '~icons/material-symbols/info-outline';
 ---
 
-<div class='bg-primary-500 !text-white p-3 my-4 rounded-lg max-w-md mx-auto box text-sm'>
-    <span style={{ fontSize: '1.0rem' }} class='!text-white'>
+<div class='bg-primary-300 p-3 my-4 rounded-lg mx-auto box text-sm'>
```
corneliusroemer, line 5: Can you show a screenshot of the changes here? What prompted you to make these changes? Just curious

monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
```diff
@@ -1,20 +1,23 @@
 ---
 layout: ../../../layouts/DocLayout.astro
 title: "Adding or removing a user group"
-order: 5
+order: 21
 ---
 
-Users cannot add themselves to groups; they must ask someone within a group to add them. When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
+Users cannot add themselves to groups; they must ask someone within a group to add them.
```
corneliusroemer, line 7:
```suggestion
Users cannot add themselves to groups; they must ask someone within a group to add them. (Users are automatically members of groups they create)

```

monorepo/website/src/pages/docs/how-to/add-remove-user-group.mdx
```diff
@@ -1,20 +1,23 @@
...
 ---
 
-Users cannot add themselves to groups; they must ask someone within a group to add them. When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
+Users cannot add themselves to groups; they must ask someone within a group to add them.
+When making this request, please be sure to supply your username exactly as it appears within your user account, as this is what they will use to add you.
 
 You can then ask them to follow the below steps to add you to their group.
 
-To add a user to your group:
+## Adding a user to your group
```
corneliusroemer, line 12:
```suggestion
## Adding a user to a group you are a member of

```

I'm not sure about this change, the wording predates this PR but thinking about it: "your group" reads more nicely but is a bit imprecise. It might give the wrong idea that there are owners of groups and that everyone has a group or belongs to one group, when in fact all members can add anyone to any groups they belong to.

monorepo/website/src/pages/docs/how-to/authentication-api.mdx
```diff
@@ -12,7 +12,7 @@ The retrieved token expires after 10 hours.
 ## Send request with cURL
 
 <InfoBox>
-    To use the <a href="https://demo.pathoplexus.org" class="!text-white">demo instance</a> instead of the main instance, please replace `authentication.pathoplexus.org` with `authentication-demo.pathoplexus.org`.
+    To use the <a href="https://demo.pathoplexus.org" class="!text-black">demo instance</a> instead of the main instance, please replace `authentication.pathoplexus.org` with `authentication-demo.pathoplexus.org`.
```
corneliusroemer, line 15: Should we maybe use a custom styled link component instead of applying the same class styling manually in 10 places? This predates the PR of course - just becomes noticeable due to the refactor in this PR.

monorepo/website/src/pages/docs/how-to/edit-group.mdx
```diff
@@ -0,0 +1,17 @@
+---
+layout: ../../../layouts/DocLayout.astro
+title: "Editing group information"
+order: 22
+---
+import CautionBox from '../../../components/asides/CautionBox.astro';
+
+Any member of a group can edit group information, which means that you can edit group information for all the groups that you are a member of.
```
corneliusroemer, line 8:
```suggestion
Any member of a group can edit group information, which means that you can edit group information for all the groups that you are a member of.

All group information except for the group's id can be edited.
```

monorepo/website/src/pages/docs/how-to/edit-group.mdx
corneliusroemer, line 16:
```suggestion
If you change the name of a group, the [`center_name`](https://ena-docs.readthedocs.io/en/latest/submit/general-guide/programmatic.html#identifying-submitters) of all future, but not past, ENA submissions will change accordingly.
```

corneliusroemer: I've added a suggestion

monorepo/website/src/pages/docs/how-to/authentication-api.mdx
theosanderson: I don't think we should use a custom styled link component. If we made any change I think it should be about new CSS classes in the mdContainer's CSS

monorepo/website/src/pages/docs/how-to/edit-group.mdx
theosanderson: (minor vote against this change which IMO would slightly reduce clarity as I don't think people would expect to be able to change a numeric ID and therefore might interpret this as being the name)

monorepo/website/src/components/asides/InfoBox.astro
theosanderson: You can see it here:
https://preview-update-docs-group-edit.pathoplexus.org/about/terms-of-use/open-data
https://pathoplexus.org/about/terms-of-use/open-data

I have slightly mixed feelings, but this probably is an improvement

monorepo/website/src/pages/docs/how-to/authentication-api.mdx
fhennig: I can also see an argument against mixing tailwind and custom CSS classes.

Another possible solution could be to add the styling to the infobox component and just style all links inside of it in a particular way, but that would also mix custom styling with tailwind.

fhennig: Example:

```astro
---
import MaterialSymbolsInfoOutline from '~icons/material-symbols/info-outline';
---

<style>
  .link-styled a {
    @apply text-red-500 underline; /* Tailwind classes */
  }

  .link-styled a:hover {
    @apply text-red-700; /* Tailwind classes for hover */
  }
</style>

<div class="bg-primary-300 p-3 my-4 rounded-lg mx-auto box text-sm link-styled">
  <span class="text-lg">
    <MaterialSymbolsInfoOutline />
  </span>
  <slot />
</div>

```

fhennig: I think this would be the easiest for the person writing the mdx file, you don't have to think about it at all when adding the link to the info box, just use the `a` tag without any stilying and it'll be fine.

corneliusroemer: I like the idea of not putting custom styling in the mdx files - this should be in component code and transparent to docs writers.

I also don't see why we should use `<a>` at all in markdown - one should use `[link text](link target)` instead. @fhennig It'd be great if you could try out your suggestion, we could see whether turning the `<a>` into `<URL>` would work.

monorepo/website/src/pages/docs/how-to/edit-group.mdx
corneliusroemer: I don't know off the top of my head - we might use both in different places - @anna-parker maybe you know it immediately?

monorepo/website/src/pages/docs/how-to/authentication-api.mdx
fhennig: it works! [ddf3c7d](https://github.com/pathoplexus/pathoplexus/pull/241/commits/ddf3c7dec3400bb5698188fddf9023abb1acc175)

corneliusroemer:
I would probably add a word "Caution" in addition to the ⚠️ sign, to not rely on signs for information conveyance
Brave Browser 2024-10-17 14 24 37

monorepo/website/src/pages/docs/how-to/authentication-api.mdx
corneliusroemer: This is really cool, thanks @fhennig!
image

@theosanderson are you happy with this?

fhennig:
Not sure if still in scope but I have nothing else to do so I'll get to it 😁

monorepo/website/src/pages/docs/how-to/edit-group.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+1. Log into your account and enter your account page by clicking 'My account' in the top-right corner of the webpage
+2. Under 'My groups' click the group that you want to edit
+3. In the group detail view, click the 'Edit group' button in the top right corner
+4. Edit any property of the group you want to change and click 'Update group' to save your changes
+
+<CautionBox>
+If you change the name of a group the center name in corresponding ENA submissions will also change.
+</CautionBox>
```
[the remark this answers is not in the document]
fhennig: Ok.

corneliusroemer (approved):
LGTM now

fhennig:
Thanks Cornelius!

@theosanderson I assume you didn't have any objections, if you do we can always make a follow-up PR but it was good to get this sorted for now.

monorepo/website/src/pages/docs/how-to/edit-group.mdx
anna-parker: Ah sorry - so currently we do not (yet) do updates in ENA submission, center name is used in the project and assembly. For new assemblies we will use the new center name but projects will (for now) keep the center name that was current when they were created. 

We should add updating in future - might not be trivial though

PR #251, opened 2024-10-22
Step 2 of author formatting changes: run v2 of prepro

anna-parker:
Test https://github.com/loculus-project/loculus/pull/2986

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1104,7 +1104,6 @@ organisms:
           nextclade_dataset_name: nextstrain/ebola/zaire
       - <<: *preprocessing
         version: 2
-        dockerTag: commit-6e05226
         configFile:
           <<: *preprocessingConfigFile
           nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1204,6 +1204,7 @@ organisms:
         defaultOrder: descending
     preprocessing:
       - <<: *preprocessing
+        replicas: 10
         configFile:
           <<: *preprocessingConfigFile
           nextclade_dataset_name: nextstrain/wnv/all-lineages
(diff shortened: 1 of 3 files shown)


loculus_values/values.yaml
```diff
@@ -1101,6 +1102,13 @@ organisms:
           <<: *preprocessingConfigFile
           nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
           nextclade_dataset_name: nextstrain/ebola/zaire
+      - <<: *preprocessing
+        version: 2
+        dockerTag: commit-6e05226
```
corneliusroemer, line 1107:
```suggestion
```

pathoplexus_app/values.yaml
```diff
@@ -1,3 +1,3 @@
-loculusVersion: 712fde0a6c1f9b8386965c470a98add9672b2ea2
+loculusVersion: 6e0522666f7fef744fb348513926c114a18a7ee9
```
corneliusroemer, line 1:
```suggestion
loculusVersion: 07f7872ce2808c45e778fa5c65f5d6989fdb7c12
```

loculus_values/values.yaml
```diff
@@ -1096,6 +1096,13 @@ organisms:
   ebola-zaire:
     <<: *defaultOrganismConfig
     preprocessing:
+      - <<: *preprocessing
+        version: 1
+        dockerTag: commit-6e05226
```
anna-parker, line 1101:
```suggestion
        dockerTag: commit-9704260
```

loculus_values/values.yaml
```diff
@@ -435,6 +435,10 @@ defaultOrganismConfig: &defaultOrganismConfig
         header: Authors
         truncateColumnDisplayTo: 15
         ingest: ncbiSubmitterNames
+        preprocessing:
```
anna-parker, line 438: sadly this means that while we are waiting for the sequences to be reprocessed the old pipeline will not be usable (throw an error due to unknown function) unless we temporarily add this to the preprocessing pipeline v1 as a no-op.

corneliusroemer: aah, good catch! dying think it's a problem since the probability of submission is low no?

anna-parker: yeah probably I created this just now as a workaround: https://github.com/loculus-project/loculus/pull/3065 but potentially we actually do want submissions to not work in the off-chance someone submits with incorrect authors format while we are doing this

PR #259, opened 2024-10-29
Update values.yaml to add more replicas

anna-parker:


corneliusroemer:
Ah I didn't see the PR did it independently

PR #261, opened 2024-10-29
Update values.yaml to v2 and enforce author format

anna-parker:


corneliusroemer:
Just for posterity (we'll forget otherwise), the following changes are happening to processed data now due to first time ever version bumping preprocessing (some changes had accumulated and were not showing up due to us not bumping):

- We now normalize booleans to `True/False`, previously they could be anything but in practice they were `true` (change from 2 months ago, now percolating)
- Frameshift ranges had a bug, could show like `199-199` or even `89-88`, now we show ranges of length 1 as `199`.

PR #274, opened 2024-11-01
Disable link checker for CI for now, still allow manual triggering

corneliusroemer:
It's too flaky

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -1,23 +1,18 @@
 name: broken-link-check
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
Thanks - to be honest my recent discussion was more about flakes like this 
image
 where it's actually (I guess) about our preview infrastructure. But I'm happy to go with this if you think it's helpful. I think eventually we should:
 - test internal links with GitHub infrastructure
 - test external links somehow potentially with a bunch of retries and long timeouts

corneliusroemer:
Ok I'll turn it off entirely in CI then but still allow it to be triggered manually as part of testing.

PR #275, opened 2024-11-01
fix: bring insdc country list up to date

corneliusroemer:
When we last changed (fixed) this on Loculus side we forgot to mirror it here. Now there are comments reminding us to check mirroring - also documenting when last changed and where to get them from.


emmahodcroft (approved):
LGTM - is same changes that went through on Loculus a while ago and were done by Anya & Cornelius

PR #276, opened 2024-11-01
Update prepro version

anna-parker:
Updating the versionComment field is equivalent to adding a new metadata field. 

This now means old sequences in the db do not have this field, but now SILO assumes they do have this field. We need to update the prepro version to give all fields this versionComment.


theosanderson (approved):
Thanks

PR #277, opened 2024-11-01
Adjust SAB order

theosanderson:
IMO it looks a bit unexpected to have a silhouette picture first?


emmahodcroft (approved):
Thanks Theo, good point!

PR #285, opened 2024-11-04
Remove unneeded fields

anna-parker:
https://github.com/loculus-project/loculus/pull/3075 makes the code dryer: we can remove code duplication in the values.yaml now.


anna-parker:
This has now been merged in https://github.com/pathoplexus/pathoplexus/pull/299

PR #289, opened 2024-11-07
Use `git clean` instead of `python delete_ignored.py`

tsibley:
No need for a custom program here.

Note that along with removing files listed in monorepo/.gitignore, the monorepo/ paths in the root dir's .gitignore will *also* be removed. This seems desirable (or at least not undesirable), but it is a slight change of behaviour.


tsibley:
Noticed this curiosity when I was reading thru some code, so figured instead of making a comment about it somewhere I'd send a patch. Take it or leave it, no hard feelings either way!

theosanderson:
Thanks @tsibley - this seems great! TIL

theosanderson (approved):

PR #332, opened 2024-12-20
Move all additional metadata descriptions and examples into config - use new MetadataTable component.

anna-parker:
This relies on https://github.com/loculus-project/loculus/pull/3477

This PR updates the metadata docs to use descriptions from the `values.yaml`. 
- Adds a desired attribute to metadata fields that we would like users to add.
- Adds definition and additional guidance fields to all metadata fields that users can input. 
- Condenses information in the guidance field to avoid repetition of information in the definition field, instead of telling users to pick options from a pick list (which is inaccurate) adds a link to ontology sites where standardized terms can be searched and chosen from.


loculus_values/values.yaml
```diff
@@ -545,6 +569,9 @@ defaultOrganismConfig: &defaultOrganismConfig
         enableSubstringSearch: true
         columnWidth: 150
         order: 50
+        desired: true
+        definition: "List of author affiliations in the same order as authors, one entry per author"
```
anna-parker, line 573: We don't have one entry per author currently - so I think this should probably be removed. 
```suggestion
        definition: "List of author affiliations"
        guidance: "If multiple affiliations exist create a semi-colon separated list in the same order as authors with one entry per author "
```

loculus_values/values.yaml
```diff
@@ -604,111 +634,113 @@ defaultOrganismConfig: &defaultOrganismConfig
...
         definition: The reason that the sample was collected.
-        guidance: Select a value from the pick list in the template
         example: Diagnostic testing
         header: Sampling
         ingest: ncbiPurposeOfSampling
       - name: presamplingActivity
         ontology_id: GENEPIO:0100433
-        definition: The activities or variables introduced upstream of sample collection that may affect the sample collected.
-        guidance: If there was presampling activity that would affect the sample prior to collection (this is different than sample processing which happens after the sample is collected), provide the experimental activities by selecting one or more values from the template pick list. If the information is unknown or cannot be provided, leave blank or provide a null value.
+        definition: Any presampling activity that would affect the sample prior to collection (this is different than sample processing which happens after the sample is collected).
```
anna-parker, line 651:
```suggestion
        definition: Any activities or variables introduced upstream of sample collection that would affect the sample prior to collection (this is different than sample processing which happens after the sample is collected).
```

loculus_values/values.yaml
```diff
@@ -604,111 +634,113 @@ defaultOrganismConfig: &defaultOrganismConfig
...
+        guidance: 'Use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/uberon. If not applicable, leave blank.'
         example: Feces [UBERON:0001988]
         displayName: Body product
         header: Sampling
       - name: environmentalMaterial
         ontology_id: GENEPIO:0001223
         definition: A substance obtained from the natural or man-made environment e.g. soil, water, sewage, door handle, bed handrail, face mask.
-        guidance: 'Provide a descriptor if an environmental material was sampled. Use the pick list provided in the template. If a desired term is missing from the pick list, use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/envo. If not applicable, leave blank.'
+        guidance: 'Use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/uberon. If not applicable, leave blank.'
```
anna-parker, line 679:
```suggestion
        guidance: 'Use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/envo. If not applicable, leave blank.'
```

loculus_values/values.yaml
```diff
@@ -604,111 +634,113 @@ defaultOrganismConfig: &defaultOrganismConfig
...
         example: Face mask [OBI:0002787]
         displayName: Environmental material
         header: Sampling
       - name: environmentalSite
         ontology_id: GENEPIO:0001232
         definition: An environmental location may describe a site in the natural or built environment e.g. hospital, wet market, bat cave.
-        guidance: 'Provide a descriptor if an environmental site was sampled. Use the pick list provided in the template. If a desired term is missing from the pick list, use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/envo. If not applicable, leave blank.'
+        guidance: 'Use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/uberon. If not applicable, leave blank.'
```
anna-parker, line 686:
```suggestion
        guidance: 'Use this look-up service to identify a standardized term: https://www.ebi.ac.uk/ols/ontologies/envo. If not applicable, leave blank.'
```

loculus_values/values.yaml
```diff
@@ -837,59 +859,60 @@ defaultOrganismConfig: &defaultOrganismConfig
...
+        definition: The length of the amplicon generated by PCR amplification including units.
         example: 300bp
         displayName: Amplicon size
         header: Sequencing
       - name: sequencingInstrument
         ontology_id: GENEPIO:0001452
         definition: The model of the sequencing instrument used.
-        guidance: Select the sequencing instrument from the pick list.
-        example: Oxford Nanopore MinION [GENEPIO:0100142]
+        # guidance: Select the sequencing instrument from the pick list.
```
anna-parker, line 878:
```suggestion
        guidance: First check if there exists a standardized term: https://www.ebi.ac.uk/ols4/ontologies/obi for your model, if not add your model in free text.
```

loculus_values/values.yaml
```diff
@@ -910,7 +932,7 @@ defaultOrganismConfig: &defaultOrganismConfig
       - name: referenceGenomeAccession
```
anna-parker, line 932: I think this field should be removed as it will potentially differ from the reference sequence we use and will be confusing

loculus_values/values.yaml
```diff
@@ -932,68 +954,78 @@ defaultOrganismConfig: &defaultOrganismConfig
...
         example: low average genome coverage
         displayName: Quality control issues
         header: Sequencing
       - name: qualityControlDetails
         ontology_id: GENEPIO:0100561
         definition: The details surrounding a low quality determination in a quality control assessment.
-        guidance: Provide notes or details regarding QC results using free text.
         example: CT value of 39. Low viral load. Low DNA concentration after amplification.
         displayName: Quality control details
         header: Diagnostics
```
anna-parker, line 1003: I also think all these values should either be removed or marked as no input as we perform additional quality control on samples and do not use QC that users upload.

PR #333, opened 2024-12-23
Remove unused metadata fields

anna-parker:
When going over the metadata fields I found a number of fields that we do not use and we should probably not allow users to input as this will confuse other users. 

These are 
- referenceGenomeAccession: We align sequences to a reference and should not allow users to specify this value themselves
-  qualityControlMethodName, qualityControlMethodVersion, qualityControlDetermination, qualityControlIssues, qualityControlDetails: we perform our own QC on sequences and show these results adding additional QC values from submitters will be confusing and is not needed

We can also condense certain fields to reduce metadata schema bloat: 
- diagnosticMeasurementValue and diagnosticMeasurementUnit can be joined
- previousInfectionOrganism 
 is not required as previousInfectionDisease is also an option and should include these details
- exposureSetting and exposureEvent can be joined

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -771,13 +771,6 @@ defaultOrganismConfig: &defaultOrganismConfig
         example: Canada, Vancouver; USA, Seattle; Italy, Milan
         displayName: Travel history
         header: Host
-      - name: exposureEvent
-        ontology_id: GENEPIO:0001417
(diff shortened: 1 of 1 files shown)


emmahodcroft:
PHA4GE info on QC tags: https://github.com/pha4ge/contextual_data_QC_tags

PR #335, opened 2024-12-23
feat(config): Generate metadata docs from values.yaml

anna-parker:
This relies on https://github.com/loculus-project/loculus/pull/3477

This PR updates the metadata docs to use descriptions from the values.yaml.

- Adds a desired attribute to metadata fields that we would like users to add.
- Adds definition and additional guidance fields to all metadata fields that users can input.
- Condenses information in the guidance field to avoid repetition of information in the definition field, instead of telling users to pick options from a pick list (which is inaccurate) adds a link to ontology sites where standardized terms can be searched and chosen from.


loculus_values/values.yaml
```diff
@@ -94,6 +94,9 @@ defaultOrganismConfig: &defaultOrganismConfig
         columnWidth: 90
         notSearchable: true
         order: 10
+        definition: "The date on which the sample was collected."
+        guidance: "Please format YYYY-MM-DD - use XX if unknown, ex: `2020-03-XX` or `20XX-XX-XX`, and provide at least year"
```
anna-parker, line 98:
```suggestion
        guidance: "Please format YYYY-MM-DD - use XX if unknown, ex: `2020-03-XX` or `2020-XX-XX`, and provide at least year"
```

emmahodcroft:
I'm happy to review this as good from my end, but I can only review the changes to the definitions/text, not the code, so dunno if there's a good way to say that a more code-y person needs to also review?

loculus_values/values.yaml
```diff
@@ -717,118 +749,109 @@ defaultOrganismConfig: &defaultOrganismConfig
...
-        guidance: The reason why a sample was originally collected may differ from the reason why it was selected for sequencing. The reason a sample was sequenced may provide information about potential biases in sequencing strategy. Provide the purpose of sequencing from the picklist in the template. The reason for sample collection should be indicated in the "purpose of sampling" field.
+        guidance: The reason why a sample was originally collected may differ from the reason why it was selected for sequencing. The reason a sample was sequenced may provide information about potential biases in sequencing strategy.
         example: Baseline surveillance (random sampling) [GENEPIO:0100005]
         displayName: Purpose of sequencing
         header: Sequencing
       - name: sequencingDate
         ontology_id: GENEPIO:0001447
         definition: The date the sample was sequenced.
-        guidance: Provide the sequencing date in ISO 8601 standard format "YYYY-MM-DD".
+        guidance: "Please format YYYY-MM-DD - use XX if unknown, ex: `2020-03-XX` or `20XX-XX-XX`, and provide at least year"
```
emmahodcroft, line 854:
```suggestion
        guidance: "Please format YYYY-MM-DD - use XX if unknown, ex: `2020-03-XX` or `2020-XX-XX`, and provide at least year"
```

monorepo/website/src/pages/docs/concepts/metadataformat.mdx
```diff
@@ -1,156 +1,29 @@
...
+## Metadata Types
 
 Some fields, like dates, countries and authors, will be standardized so that all entries in the database are easy to process.
 At the moment, these are the only fields we standardize, and we will keep users updated if that changes in the future.
 
+### Dates
+
 {/* Please format dates as YYYY-MM-DD - use XX if unknown, ex: `2020-03-XX` or `20XX-XX-XX`. */}
 Please format dates as YYYY-MM-DD.
 To submit sequences, the full date (including day) is required. If you have older sequences for which the full date isn't available, or legal reasons relating to releasing the full date, please get in touch via [submission@pathoplexus.org](mailto:submission@pathoplexus.org).
```
emmahodcroft, line 21: Is this still true? And I guess if it isn't technically, do we want to keep saying this to encourage it, or is it just miseading if we actually allow incomplete dates? (Which I think we do currently?)

emmahodcroft:
If you ping me (on slack) when you've added those few comments above @anna-parker I can approve :)

PR #341, opened 2025-01-21
feat: Enable earliest release date

fhennig:
The earliest release date feature was added to loculus in this PR: https://github.com/loculus-project/loculus/pull/3380/files#diff-8c45270b76bf36f08f571b509614d6672984b0604844a487a0ad3ba2ad794943

But to use it, changes to the values.yaml need to be made.

PR #344, opened 2025-01-21
Revert "feat: Enable earliest release date"

theosanderson:
Reverts pathoplexus/pathoplexus#341

As Cornelius has mentioned - we may want to rename this field before enabling it


corneliusroemer:
Thanks! This is good but we might also want to revert the PR on Loculus as I'm not sure we can rename that field just in config. If the change needs to be made on Loculus, we might as well revert it there as well?

theosanderson:
To me it seems simpler to just rename it in Loculus

PR #351, opened 2025-02-03
Add metadata template for esud

theosanderson:
https://preview-add-metadata-template-for-esud.pathoplexus.org/

I tested and got:

[Ebola_Sudan_metadata_template.tsv.fixed.txt](https://github.com/user-attachments/files/18642138/Ebola_Sudan_metadata_template.tsv.fixed.txt)

[I originally uploaded the wrong file here, sorry]

which looks good to me

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1262,6 +1262,25 @@ organisms:
       <<: *schema
       organismName: "Ebola Sudan"
       image: "/images/organisms/ebolasudan_small.jpg"
+      metadataTemplate: 
+        - geoLocCountry  
+        - sampleCollectionDate  
+        - submissionId  
+        - authorAffiliations  
+        - authors  
+        - collectionDevice  
+        - collectionMethod  
+        - cultureId  
+        - depthOfCoverage  
+        - geoLocAdmin1  
+        - geoLocAdmin2  
+        - hostNameScientific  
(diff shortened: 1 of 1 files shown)


anna-parker:
@theosanderson I am a bit confused the template you linked to contains additional, not desired fields? Also I can't see these changes here: 
image

theosanderson:
Sorry @anna-parker I uploaded the wrong file to the PR description before

anna-parker:
I wonder if we should somehow specify that this is not a list of all fields. I just wonder if it might be confusing to have different fields in this template and in the linked docs. But otherwise this looks great - I am ok with approving as is :-)

anna-parker (approved):

emmahodcroft:
It would be nice (not necessarily in this PR) to clarify that one can find a list of all fields.. (not necessarily for this PR - could be separate issue)

Could change wording from:
`The documentation pages contain more details on the required [metadata format (https://staging.pathoplexus.org/docs/concepts/metadataformat).`
to
`The documentation pages contain more details on the required [metadata format](https://staging.pathoplexus.org/docs/concepts/metadataformat), including a list of all supported metadata fields.`

theosanderson:
Gonna merge this to ensure it gets in - happy with various other things downstream. Also this only applies to one organism for now!

PR #364, opened 2025-02-04
Apply metadata template to all organisms

theosanderson:
In #351 we added a shorter metadata template for Ebola sudan. Here we instead apply this same template across all Pathoplexus organisms.

This also slights reorders the template. 

I've tested that it renders as expected across organisms.

http://preview-template.pathoplexus.org


anna-parker (approved):
I just tested and LGTM!

theosanderson:
Thanks!

PR #367, opened 2025-02-06
feat(website): use Pathoplexus' own metadata docs page

chaoran-chen:
This is to configure the website to keep using Pathoplexus' own metadata docs page after https://github.com/loculus-project/loculus/pull/3634.


corneliusroemer:
Cherry picked onto #374

PR #380, opened 2025-02-12
Boost memory for LAPIS and SILO in prod

theosanderson:
Mpox is crashing in prod with sequence downloads. This PR:

- boosts the request memory (i.e. the amount reserved)
- decreases the limit to be the same as the requested amount so that things are not determined by outside availability which could mask problems


corneliusroemer:
That's never gonna work because many SILOs, e.g. mix require at least 2.7GB of RAM

theosanderson:
Yep Chaoran made same point

PR #385, opened 2025-02-12
feat(deployment): Tune LAPIS/SILO requests per organism

corneliusroemer:


corneliusroemer:
It remains high for everything but production (which overrides these default).

I'm not sure we have _that_ much memory: we have 6x32GB, and each instance would then require ~15GB for mpox alone. With main, staging, production, demo, we'd use up 60GB of the 180GB.

The 5GB for Silo is 2x what it actually needs, should only get killed after some time or 3 full downloads which is fine for previews etc I think.

theosanderson:
Sorry, you are right I hadn't realised you weren't targeting production. Hmm.. for previews I think having a difference between the `limit` and the `requests` is quite good, so you basically take what you can get?

theosanderson:
But yeah, I guess this is well less than one node's RAM - although maybe we should be trying to cut down our number of nodes a bit?

theosanderson:
ah - although with that for each preview - that does add up

corneliusroemer:
I've reduced requests to be more reflective of the actual usage, while having a bit of buffer in limits so we don't get OOM too often (except for SILO which has that memory leak). Does that look better?

theosanderson:
Cool, yeah I think these look reasonable. I think there's a last thing that we should likely do anyway which is to set production-specific versions of these that equalise the `limit` and the `requests` to avoid OOM-killing due to other deployments. I'm not sure if we need to block on that: the reason we might want to would be that here by playing with the `requests` we might make problems more likely in production, but it might also be fine. If you can be bothered to add it, great, if not I can probably still approve

Issue #395, opened 2025-02-18
[FEAT] make `isLabHost` a proper boolean field


fengelniederhammer:
**Is your feature request related to a problem? Please describe.**

`isLabHost` has more values than I would expect:

```
$ curl 'https://lapis.pathoplexus.org/west-nile/sample/aggregated?fields=isLabHost' --silent | jq
{
  "data": [
    {
      "count": 20811,
      "isLabHost": null
    },
    {
      "count": 126,
      "isLabHost": "true"
    },
    {
      "count": 63,
      "isLabHost": "True"
    },
    {
      "count": 630,
      "isLabHost": "False"
    }
  ],
  "info": ...
}
``` 

This is for West Nile, but I assume that it's also the case for other organisms.

**Describe the solution you'd like**

SILO allows boolean fields. It would be best to use that. The preprocessing pipeline needs to make sure that the values are cleaned properly.

**Describe alternatives you've considered**

**Additional context**

PR #396, opened 2025-02-18
feat!(config): Configure isLabHost as a boolean

anna-parker:
Resolves: https://github.com/pathoplexus/pathoplexus/issues/395

This will require bumping the prepro pipeline

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1095,6 +1095,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         ingest: ncbiIsLabHost
         definition: "If a laboratory host (e.g. cultured cell line) was used to propagate the sample."
         example: "True"
+        type: boolean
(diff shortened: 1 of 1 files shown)


theosanderson:
(nit: I think of "assert" as having a specific meaning that's a bit different to this which is why I renamed the PR: https://chatgpt.com/share/67c59f73-44d0-8005-89ca-5a40d3ab37e3)

theosanderson (approved):

theosanderson:
Will this make new versions of everything? If so good to document in the PR description so we know when deploying

anna-parker:
> Will this make new versions of everything? If so good to document in the PR description so we know when deploying
It shouldn't make new versions as in loc_xx.2 - this will require reprocessing of the current data by increasing the pipeline version - so that all the data is reprocessed - so yes there will be a new processing version

PR #404, opened 2025-02-26
Add some default download fields

theosanderson:
This is just a starting point - I don't think we need to agonise too much over this as we can easily change them later

Only relevant after [3761](https://github.com/loculus-project/loculus/pull/3761)


fhennig (approved):
I'm happy with the selection for now - I'm approving mostly so we have something for when we merge the other PR.

If more PPX-affiliated people would like to chime in, please do!

Issue #407, opened 2025-03-03
[BUG] preview-main.pathoplexus is not showing any sequences


anna-parker:
I was just going to rollout some new changes and saw that Pathoplexus main has now sequences, looking into the logs this is because LAPIS/SILO is erroring in silo-preprocessing with:
```No old input data found at /preprocessing/input//data.ndjson.zst
No LINEAGE_DEFINITIONS given, nothing to configure;
Starting preprocessing
[2025-03-03 13:21:36.371] [logger] [info] [main.cpp:95] Starting SILO (version 0.5.3)
[2025-03-03 13:21:36.371] [logger] [info] [environment_variables.cpp:110] Given env variable 'SILO_DEFAULT_RUNTIME_CONFIG' is not a valid key for 'silo preprocessing' but in allow_list.
[2025-03-03 13:21:36.372] [logger] [info] [main.cpp:117] preprocessing_config = {"database_config_file":"database_config.yaml","duckdb_memory_limit_in_g":null,"input_directory":"/preprocessing/input/","intermediate_results_directory":"/preprocessing/temp/","lineage_definitions_file":null,"ndjson_input_filename":"data.ndjson.zst","output_directory":"/preprocessing/output/","preprocessing_database_location":null,"reference_genome_file":"reference_genomes.json"}
[2025-03-03 13:21:36.372] [logger] [info] [database_config.cpp:214] Reading database config from /preprocessing/input/database_config.yaml
[2025-03-03 13:21:36.374] [logger] [error] [main.cpp:52] Internal Error: Metadata 'isLabHost' generateIndex is set, but generating an index is only allowed for types STRING
```
The issue seems to be having `SILO_DEFAULT_RUNTIME_CONFIG` in the config.

anna-parker:
Ah no this is linked to the `isLabHost` change: I need to remove `generateIndex: true` - we don't use `generateIndex: true` in Loculus so that is how this wasn't noticed.

theosanderson:
Might be nice to see if there's a way to put this dependency in the schema

PR #409, opened 2025-03-03
fix(config): remove generateIndex for boolean values as currently not supported by SILO

anna-parker:
Resolves: https://github.com/pathoplexus/pathoplexus/issues/407

Preview: https://preview-fix-boolean.pathoplexus.org/


theosanderson:
(there's some noise in the diff?)

corneliusroemer (approved):

PR #423, opened 2025-03-14
Revert "increase lapis memory for mpox in prod"

anna-parker:
Reverts pathoplexus/pathoplexus#419


theosanderson (approved):

theosanderson (approved):
We may want to increase `requests` for prod at some point, but since it was OK before it's probably OK as a starting point!

PR #442, opened 2025-04-25
Update mpox dataset version to 2025-04-25

corneliusroemer:
Standard dataset update: adding new prepro version (9 in this case), keeping old one as is. Once upgrade is completed we can remove version 8 list item.

There are no clade/lineage changes between old and new version so differences should be small.

Using a yaml anchor & to avoid having to repeat the genes and keep the diff small.

Preview: https://preview-update-mpox-dataset.pathoplexus.org/

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1599,8 +1599,9 @@ organisms:
         defaultOrder: descending
     preprocessing:
       - <<: *preprocessing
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
LGTM! Do you want to do this on staging already or wait for the other rollout on Monday?

PR #453, opened 2025-05-09
Add display names to more fields

theosanderson:
Add display names for more fields

https://preview-add-display-names.pathoplexus.org/

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1482,15 +1482,16 @@ organisms:
       organismName: "Crimean-Congo Hemorrhagic Fever Virus"
       nucleotideSequences: [L, M, S]
       image: "/images/organisms/cchf_small.jpg"
+      metadataAdd:
         - name: totalStopCodons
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -1462,19 +1483,15 @@ organisms:
       nucleotideSequences: [L, M, S]
       image: "/images/organisms/cchf_small.jpg"
       metadataAdd:
-        - name: hostNameScientific
-          generateIndex: true
-          autocomplete: true
-          header: "Host"
-          ingest: ncbiHostName
-          initiallyVisible: true
```
theosanderson, line 1470: This seemed to be a dupe?

anna-parker (approved):
Thanks for this!

I also just now had a look through and 
- I noticed `sequencedByOrganization` has displayName name: `sequenced by` maybe we should change to `sequenced by (organization)`

- `diagnosticTargetGeneName` shoudl maybe be `Diagnostic Target Gene Name`

PR #456, opened 2025-05-14
request CPU for all organism-specific, prod-specific resource specifications

theosanderson:


theosanderson:
I set these limits relatively consistently with the other `silo` specification - I don't think we ever intended to overwrite and remove that we just did accidentally

corneliusroemer:
Good catch @theosanderson, that was probably my mistake - if we did yaml referencing/inheritance we could condense this a bit

theosanderson:
It was possibly both of us I think :)

PR #463, opened 2025-05-21
Add linkouts

anna-parker:


chaoran-chen (approved):

corneliusroemer:
Was it intentional to leave out max seqs for most pathogens, e.g. RSV etc?

PR #466, opened 2025-05-25
feat: add script for revoking suppressed sequences

anna-parker:
Add a script to automate revoking suppressed sequences when we receive alerts such as https://loculus.slack.com/archives/C08CLH54PFD/p1748184967120199 - it also now double checks that sequences are indeed suppressed before revoking them.


corneliusroemer (commented):
The script could be a lot shorter if we used code from ingest - I wonder if we could create a pip library or so to reuse code from?

Duplicating once is fine I guess, but we might be getting to a point where a Loculus client python library might make sense

anna-parker:
> The script could be a lot shorter if we used code from ingest - I wonder if we could create a pip library or so to reuse code from?

In future we might actually want ingest to run these commands automatically when it encounters potentially suppressed sequences - so I think the duplication is ok for the moment - but yes creating a library for commonly used ingest commands (there is also some overlap with functions in prepro and ena deposition) might be useful!

PR #470, opened 2025-05-27
DO NOT MERGE: Test new revoke and regroup changes 

anna-parker:
point to https://github.com/loculus-project/loculus/pull/4177 (commit before switching to new CCHF dataset)


anna-parker:
already tested :-)

PR #476, opened 2025-05-28
Reduce `defaultResources`, and backend and ingest resources for non-prod, non-staging deployments

theosanderson:
Reduces these resources so that more non-production things can fit into the resources that aren't reserved by production (and these can also use resources reserved by production but not used)


loculus_values/values.yaml
```diff
@@ -2438,3 +2438,25 @@ organisms:
...
+    cpu: "10m"
+  limits:
+    memory: "1Gi"
+    cpu: "1000m"
+
+resources:
+  ingest:
+    requests:
+      memory: "1Gi"
+      cpu: "50m"
```
theosanderson, line 2453: We could also alternatively this in Loculus itself (is `200m`)

anna-parker: Ingest is often using the full 100m:
image

theosanderson: Feel free to propose changes - I just copied this over from the Loculus values.yaml and reduced the CPU

loculus_values/environment_specific_values/staging.yaml
```diff
@@ -64,4 +64,11 @@ seqSets:
...
-    issueTemplate: sequence-metadata-issue.md
\ No newline at end of file
+    issueTemplate: sequence-metadata-issue.md
+defaultResources:
+  requests:
+    memory: "200Mi"
+    cpu: "20m"
+  limits:
+    memory: "1Gi"
+    cpu: "20m"
```
corneliusroemer, line 74:
```suggestion
    cpu: "20m"

```

anna-parker, line 74: These are the current loculus default values

loculus_values/values.yaml
```diff
@@ -2438,3 +2438,25 @@ organisms:
...
+    limits:
+      cpu: "50m"
+      memory: "10Gi"
+  backend:
+    requests:
+      memory: "640Mi"
+      cpu: "100m"
+    limits:
+      memory: "3Gi" # Backend requires at least 635741K of memory
```
corneliusroemer, line 2462:
```
  backend:
    requests:
      memory: "640Mi"
      cpu: "250m"
    limits:
      memory: "3Gi" # Backend requires at least 635741K of memory
```

loculus_values/environment_specific_values/staging.yaml
corneliusroemer: What's the reason you want to specify different resources for staging specifically? 

What's the logic behind:
- Slightly higher cpu requests (20 vs 10)
- But much lower cpu limits (20 vs 1000)
vs what's in values.yaml (and hence the setting for all non-prod namespaces unless they overwrite differently - none of them do except for prod)

theosanderson: It just felt conservative to keep staging exactly as it was before my PR - which is what this does because previously it inherited these from the base Loculus `values.yaml` but I don't feel strongly - would also be happy to remove this special config!

loculus_values/values.yaml
theosanderson: > I would use the same limits from loculus then actually if thats ok

I'm not sure if you are proposing here that we leave out the `limit` - if we did I'm not certain that the limit from the original Loculus values would come through and be combined (this is the problem we had that prompted https://github.com/pathoplexus/pathoplexus/pull/456)

PR #481, opened 2025-05-30
Add GitHub workflow to check helm templating works correctly

theosanderson:
This PR adds a GitHub Actions workflow that validates our Helm templating for all defined environments (staging, production, demo, main). The workflow ensures that our Helm charts can be successfully templated with the appropriate values for each environment.

### Changes
- Added `.github/workflows/helm-template-check.yml` workflow that:
  - Dynamically discovers all environments from `loculus_values/environment_specific_values/`
  - Clones the Loculus repository at the version specified in `pathoplexus_app/values.yaml`
  - Runs `helm template` for each environment with the correct values hierarchy
  - Provides debug output if templating fails

### Why is this needed?
Currently, we don't have automated checks to ensure our Helm configurations are valid. This could lead to deployment failures that are only discovered when attempting to deploy to an environment. This CI check will catch configuration errors earlier in the development process.

### How it works
1. The workflow triggers on PRs and pushes to main when relevant files change
2. It first discovers all available environments by listing YAML files in `environment_specific_values/`
3. For each environment, it:
   - Checks out both the Pathoplexus and Loculus repositories
   - Runs `helm template` with the three-layer values hierarchy:
     - Base Loculus values
     - Pathoplexus common values
     - Environment-specific values
4. If templating fails, it re-runs with debug output to help diagnose the issue

### DIFF
--- a/.github/workflows/helm-template-check.yml
+++ b/.github/workflows/helm-template-check.yml
@@ -1,0 +1,87 @@
+name: Helm Template Check
+
+on:
+  pull_request:
+    paths:
+      - 'loculus_values/**'
+      - 'pathoplexus_app/**'
+      - '.github/workflows/helm-template-check.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'loculus_values/**'
+      - 'pathoplexus_app/**'
+      - '.github/workflows/helm-template-check.yml'
+
+jobs:
+  get-environments:
+    runs-on: ubuntu-latest
+    outputs:
+      environments: ${{ steps.get-envs.outputs.environments }}
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-template-check.yml
```diff
@@ -0,0 +1,87 @@
+name: Helm Template Check
+
+on:
+  pull_request:
+    paths:
```
corneliusroemer, line 5: We should also run when loculus commit pointed at changes

theosanderson: that's in pathoplexus_app?

corneliusroemer (approved):

PR #514, opened 2025-06-10
Amend FAQ

theosanderson:


corneliusroemer (approved):
Oh yeah that was confusingly worded

PR #515, opened 2025-06-10
Improve website and dev docs

corneliusroemer:
## Summary
- link to Data Use Terms from Pathoplexus Values
- fix typos in Executive Board guidelines
- correct Zurich spelling in funding instructions

------
https://chatgpt.com/codex/tasks/task_e_684874592a2c8325b1d63b38ec164641


corneliusroemer:
German would be Zürich - Zuerich would be transliteration of ü -> ue but as there's an English standard spelling we should use that

chaoran-chen:
"Zuerich" is what Wise shows for the corresponding BIC, so I would guess that's the official value in international banking? https://wise.com/gb/swift-codes/bic-swift-code-checker?code=UBSWCHZH80A - but I'm sure that "Zurich" is also fine.

PR #535, opened 2025-06-16
Fix RSV Nextclade dataset version

corneliusroemer:


loculus_values/values.yaml
```diff
@@ -2266,6 +2266,7 @@ organisms:
         configFile:
           <<: *preprocessingConfigFile
           nextclade_dataset_name: "nextstrain/rsv/a/EPI_ISL_412866"
+          nextclade_dataset_version: "2024-11-27--02-51-00Z"
```
anna-parker, line 2269: Where is this used in loculus? I can't see it in values.yaml or prepro config?

corneliusroemer: You're right it isn't!

anna-parker: so it should be `nextclade_dataset_tag` instead of `nextclade_dataset_version`?

corneliusroemer: yep

anna-parker:
Can we also tag hmpv as we use an official release there as well?

corneliusroemer:
Added hmpv as this is also a public tagged dateset

PR #536, opened 2025-06-16
Fix: don't fake-complete incomplete dates in displayName

corneliusroemer:
- **Add mpox lineage hierarchies**
- **Add RSV-A/B lineage system definitions**
- **Fix displayName using complete dates**


anna-parker (approved):
lgtm

PR #542, opened 2025-06-23
feat: add hmpv lineage hierarchy to allow wildcard queries

corneliusroemer:
https://preview-hmpv-hierarchy.ppx.bio

image


anna-parker (approved):
Amazing! The conversion for these lineages is quite straight forward but I wonder if we could add a script to https://github.com/pathoplexus/silo-lineage-hierarchy-definitions/blob/main/definitions/hmpv/lineages.yaml to generate the file directly from https://github.com/nextstrain/hmpv/blob/main/nextclade/resources/clades.tsv in order to update it when/if there are new clades - but not required for this PR

PR #548, opened 2025-07-02
feat: Make authors required

### LINKED ISSUE #541: make `authors` a required field
If we do not submit an assembly with the AUTHORS field filled out ENA will set the broker account owner as the author (this happened e.g. for the RSV sample where Emma was set as the authors). We never want to be set as the author.

We should set authors as a required field on Pathoplexus to never get into this situation.

fhennig:
resolves #541

change `authors` from desired to required

preview: https://preview-make-authors-required.pathoplexus.org

### Screenshot

image

Authors now appears under the required fields.


anna-parker (commented):
Im pretty sure this needs some docs changes - you can search the ppx docs for required fields. We should add more info on the metadata page (you can copy the guidance there)

Issue #547, opened 2025-07-02
[FEAT] Add comment / notes field


chaoran-chen:
A general field for `comment` / `notes` should be added where the submitter can enter arbitrary text information.

anna-parker:
This means creating a new metadata field in the values.yaml

PR #566, opened 2025-07-09
feat(config): add comment field

anna-parker:
resolves: https://github.com/pathoplexus/pathoplexus/issues/547

copy of https://github.com/pathoplexus/pathoplexus/pull/563

As discussed in https://loculus.slack.com/archives/C05G172HL6L/p1751448536534509?thread_ts=1751446873.936119&cid=C05G172HL6L we should call this field comment

preview: https://preview-anya-notes-field.pathoplexus.org/

image
image

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1279,7 +1279,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           inputs: {input: nextclade.coverage}
       - name: comment
         displayName: Comment
-        header: "Additional information"
+        header: "Submission details"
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -1277,6 +1277,10 @@ defaultOrganismConfig: &defaultOrganismConfig
           type: percentage
         preprocessing:
           inputs: {input: nextclade.coverage}
+      - name: comments
+        displayName: Comments
+        header: "Additional information"
```
anna-parker, line 1282: Maybe we should put this under `Other`? What do you think @theosanderson

PR #592, opened 2025-07-24
Add Measles virus resolution

emmahodcroft:
EB has voted to add measles to PPX - this is adding the resolution that shows this

I would suggest as we've generally done in the past that we don't merge this in until we add measles - once we have a branch that adds measles, we could add this to that, if easier. 

Please double check the PDF works for you, and the link works for you


emmahodcroft:
@theosanderson this should also be merged in as part of today's launch :) (Sorry if already on your radar!)

chaoran-chen (approved):

PR #597, opened 2025-07-25
Remove v10 prepros as all organisms are now on v11

corneliusroemer:


loculus_values/values.yaml
```diff
@@ -1336,7 +1336,7 @@ defaultOrganismConfig: &defaultOrganismConfig
   preprocessing:
     - &preprocessing
       replicas: 1
-      version: 10
+      version: 12
```
anna-parker, line 1339: this should stay at 11 no?

anna-parker: I updated ;-)

PR #643, opened 2025-09-10
Tune requests recreated

corneliusroemer:


theosanderson:
This would significantly decrease number of previews possible right?

corneliusroemer:
I'm not sure - this is just reviving the old PR that couldn't be reopened easily due to being rebased #385

PR #644, opened 2025-09-10
Drop unused fields

anna-parker:
When going over the metadata fields I found a number of fields that we do not use and we should probably not allow users to input as this will confuse other users. 

These are 
- `referenceGenomeAccession`: We align sequences to a reference and should not allow users to specify this value themselves
-  `qualityControlMethodName`, `qualityControlMethodVersion`, `qualityControlDetermination`, `qualityControlIssues`, `qualityControlDetails`: we perform our own QC on sequences and show these results adding additional QC values from submitters will be confusing and is not needed

We can also condense certain fields to reduce metadata schema bloat: 
- `diagnosticMeasurementValue` and `diagnosticMeasurementUnit` can be joined
- `previousInfectionOrganism` 
 is not required as `previousInfectionDisease` is also an option and should include these details
- `exposureSetting` and `exposureEvent` can be joined

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -957,7 +957,6 @@ defaultOrganismConfig: &defaultOrganismConfig
       - name: exposureSetting
         ontology_id: GENEPIO:0001428
         definition: The setting or event leading to exposure.
-        guidance: Select the host exposure setting(s) from the pick list provided in the template. If a desired term is missing, contact the curation team.
         example: Healthcare Setting [GENEPIO:0100201]
         displayName: Exposure setting
         header: Host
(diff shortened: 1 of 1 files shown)


theosanderson:
And just a general reminder that the reason we have these as is from the PHA4GE contextual schema - which doesn't necessarily mean we can't change but just worth bearing in mind as they have been thought through by someone. For example the `diagnosticMeasurement` value would be a numeric value. (I'm not against changing things and I do think there are some things that don't make sense in our context atm as it imagines a more nested structure than we support - potentially we might even want JSON for some complex fields, though that's messy).

PR #653, opened 2025-09-15
Revert "Bump RSV-B dataset and remove obsolete mpox v11"

anna-parker:
Reverts pathoplexus/pathoplexus#650 - lets clean up mpox in a different PR


corneliusroemer:
We should not revert mpox

corneliusroemer (changes requested):
Mpox changes should not be reverted

PR #695, opened 2025-10-06
New design for Pathoplexus homepage and associated pages

theosanderson:
To unlock the new navigation https://github.com/loculus-project/loculus/pull/5085 and for other reasons we have redesigned the Pathoplexus homepage to:

- surface a big Submit button
- surface the latest news article

image

The organism selector page has been removed from Loculus due to the redesign but here a version is added back to Pathoplexus to support these buttons.

We should rename Crimean Congo.. to CCHF sometime which will improve things a bit but would prefer to do that later after merging a PR to provide full name on mouseover

https://preview-cta2.pathoplexus.org/


monorepo/website/src/components/CryptoDonationWidget.tsx
theosanderson, line 1: Sorry, these changes are just `npm format`ing this file for the first time

pathoplexus_app/values.yaml
```diff
@@ -1,3 +1,3 @@
-loculusVersion: b6cf4bea0fee2457740a37d98b0c44588d928000
+loculusVersion: eb0a110ffdf0fd21aad1c43cdc6be346b0a6268f
```
theosanderson, line 1: Revert before submission

monorepo/website/src/layouts/DocLayout.astro
theosanderson, line 1: Oops - this shouldn't have been checked in - I should remove this file and just use the one from Loculus

monorepo/website/src/layouts/AboutLayout.astro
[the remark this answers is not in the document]
theosanderson: (will fix)

monorepo/website/src/pages/index.astro
```diff
@@ -0,0 +1,142 @@
...
+                latestNewsItem && (
+                    <aside class='flex h-full flex-col justify-between rounded-2xl border border-base-200/70 bg-base-100 p-6 shadow-[0_4px_12px_rgba(0,0,0,0.05)]'>
+                        <div>
+                            <span class='text-xs font-semibold uppercase tracking-wide text-primary/80'>Latest news</span>
+                            <h2 class='mt-3 text-xl font-semibold leading-7 text-slate-900'>
+                                <a class='hover:underline' href={`/news/${latestNewsItem.slug}`}>
+                                    {latestNewsItem.title}
+                                </a>
+                            </h2>
+                            <div class='mt-1 text-sm text-slate-500'>By the Pathoplexus Team – {latestNewsItem.date}</div>
```
anna-parker, line 102: For consistency we should probably get the `websiteName` here instead of writing pathoplexus

theosanderson: (Thanks for the review!) I don't think this is a problem (if anything I'd probably hardcode pathoplexus throughout in the pathoplexus repo, but I don't think it's worth the time cost of going through and doing so)

anna-parker (approved):
Awesome thanks!

PR #709, opened 2025-10-21
test(integrity): replace sponge and use TSV-ESCAPED dataformat

corneliusroemer:
- **fix(test): replace sponge with temp files, as sponge struggles with >2GB on macOS**
- **use tsv-escaped for metadata download**

Also speeds up a lot by writing intermediary files as zstd compressed to disk.


corneliusroemer:
Verbally approved by @anna-parker

PR #769, opened 2025-12-11
Add GA minutes 20 Nov 2025 meeting

emmahodcroft:
Adds summary of minutes from GA meeting on 20 Nov 2025 to website - very small change.

Checked locally, please do check in preview:
https://preview-ga-mins-nov.pathoplexus.org/about/governance/minutes

- [x] Shows up fine on minutes page
- [x] link to pdf works


anna-parker:
Also a typo it should be `statutes` not `Statues` - this is in the text a couple times and should probably be corrected. Also I think `statutes` should be lowercase. 

Also the line `We do have the option to have non-voting members in the Statues` - I think it should be rephrased as `According to the statutes we have the option to have non-voting members`

anna-parker (commented):
PDF link looks good _ I found one type that we should probably fix before merging :-)

PR #770, opened 2025-12-11
refactor(regression-tests): adapt snakemake file to allow easy change of keyField when comparing across previews

anna-parker:


data-integrity-tests/regression-testing/Snakefile
```diff
@@ -6,6 +6,11 @@ wildcard_constraints:
     difftype=r"seq|meta",
 
 
+# The primary key field used to join entries that should be compared.
+# In production and staging this is the accessionVersion, however to compare across previews it should be changed to submissionId.
+keyField = "accessionVersion"
```
corneliusroemer: You could read this from config - then it's easy to override dynamically when invoking with `--config, -C [KEY=VALUE ...]`

corneliusroemer (approved):

PR #779, opened 2025-12-15
Revert "Update Loculus version to 012694, configure silo hard refresh, prepro upgrade intervals, zstd compression level"

anna-parker:
Reverts pathoplexus/pathoplexus#777


anna-parker:
I shoudln't have merged in the previous PR as I didnt resolve all breaking config changes- lets revert and do this correctly

PR #795, opened 2026-01-08
feat(website): add legal notice page

chaoran-chen:
This adds a legal notice page as discussed on Slack.

Preview: https://preview-legal-notice.pathoplexus.org/about/governance/legal-notice

### DIFF
--- a/monorepo/website/src/pages/about/governance/legal-notice.mdx
+++ b/monorepo/website/src/pages/about/governance/legal-notice.mdx
@@ -12,7 +12,7 @@ Switzerland
 
 **UID:** CHE-296.387.244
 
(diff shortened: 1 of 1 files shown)


monorepo/website/src/pages/about/legal-notice.mdx
theosanderson: My instinct would be to put this under Governance in the menu given that it is already included in the footer.

(I still think it's a kind of confusing page as I expect it from the name to maybe say "Notice is hereby given that..." or something, so my motivation is basically making it less prominent given that it's already prominent enough by being linked from every page).

theosanderson (approved):

monorepo/website/src/pages/about/governance/legal-notice.mdx
chaoran-chen: Sure, that's alright with me, I moved it.

PR #803, opened 2026-01-27
Change measles ingest taxon_id from 11234 to 3052345

### LINKED ISSUE #802: Measles taxon_id 11234 vs 3052345
Hi PPX team,

I just learned from @DOH-DAH0303 that public health labs submit measles samples to NCBI under taxon id 3052345 (morbillivirus hominus) instead of the current taxon id 11234 used by [PPX ingest](https://github.com/pathoplexus/pathoplexus/blob/bf62637ae7392f3edb897475835725ae7597d0d1/loculus_values/values.yaml#L2866-L2870). NCBI will eventually move records from 3052345 to 11234, but there may be a delay of a few weeks.

Thoughts on switching the measles ingest taxon id to 3052345 to get all records earlier?

theosanderson:
Updated taxon_id for ingest configuration.

resolves #802

https://preview-measles-ingest.pathoplexus.org/measles/search?

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2867,7 +2867,7 @@ organisms:
       <<: *ingest
       configFile:
         <<: *defaultIngestConfigFile
-        taxon_id: 11234
+        taxon_id: 3052345
     enaDeposition:
       singleReference:
         configFile:


loculus_values/values.yaml
```diff
@@ -2867,7 +2867,7 @@ organisms:
       <<: *ingest
       configFile:
         <<: *defaultIngestConfigFile
-        taxon_id: 11234
+        taxon_id: 3052345
```
anna-parker: we need to add this to the loculus mirror first :-)

PR #891, opened 2026-04-10
fix: Make hostTaxonId required for dengue and yellow fever

maverbiest:
Making `hostTaxonId` a required field for dengue and yellow-fever so that submissions without `host` are rejected


anna-parker (approved):
awesome!

PR #896, opened 2026-04-13
fix: Set maxNumberOfRecommendedEntries: 45000 for all mapoplexus linkouts

maverbiest:
The linkout to Mapoplexus for does not work when it is called with all ~57k dengue sequences:

image

However it does work for:
- All of DENV-1 (23k sequences)
- All of DENV-2 (20K sequences)
- All of DENV-3 (10k sequences)
- All of DENV-4 (5k sequences)
- Two seqs of each serotype
- All of RSV-A (41k sequences)

Potentially, this points to dengue having too many sequences to show in Mapoplexus at once. 

To address this for now, we will set `maxNumberOfRecommendedEntries` to 45000 for all Mapoplexus linkouts so users see a warning before attempting to use Mapoplexus with a large amount of sequences.

### Manual testing
I set up a preview and waited until more than 45000 dengue sequences were available. When I tried to send all of these to Mapoplexus via the linkout, I was met with this warning:

image

preview: https://preview-n-seqs-mapoplexus.pathoplexus.org/


anna-parker (approved):
awesome!

PR #913, opened 2026-05-01
Automatically add preview links to PRs when label added

theosanderson-agent:
Matches Loculus behaviour with automatic preview links :)

🚀 Preview: https://preview-pr-previews.pathoplexus.org

### DIFF
--- a/.github/workflows/add-preview-url.yml
+++ b/.github/workflows/add-preview-url.yml
@@ -1,0 +1,42 @@
+name: Add Preview URL to PR
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  update-pr-description:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update PR description with preview URL
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+            const hasPreviewLabel = pr.labels.some((label) => label.name === 'preview');
+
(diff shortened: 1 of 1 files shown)


.github/workflows/add-preview-url.yml
```diff
@@ -0,0 +1,42 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+  update-pr-description:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update PR description with preview URL
+        uses: actions/github-script@v8
```
anna-parker: we are one version higher on loculus
```suggestion
        uses: actions/github-script@v9
```

.github/workflows/add-preview-url.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
```
anna-parker, line 16: @theosanderson I think this requires a github token to run as it writes to the repo? we have one on loculus
```
github-token: ${{ secrets.LOCULUS_BOT_WRITE_PRS }}
```

theosanderson: It seems to work as is :D (I had it on `pull_request_target` before so it wasn't actually running against this PR, but I changed it to `pull_request` to allow testing, and it's probably fine like that anyhow)

theosanderson: Thanks for checking tho, maybe we do something unnecessary on loculus, not sure.

anna-parker: ah ok thats cool! then happy to approve :-)

anna-parker (approved):

PR #919, opened 2026-05-05
feat(deployments): reduce defaultResources limits and ingest requests (for non prod/staging)

anna-parker:
Note that staging and prod overwrite the `defaultResources` and the ingest resources are overwritten by production. These changes will only affect previews, main, the demo and staging ingest requests. Ingest pods will use the limit if it is available (this causes the memory peaks - see below).

Memory peaks for mpox on main:
image

mpox memor with lower limit:
image

🚀 Preview: https://preview-limits.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -3633,7 +3633,7 @@ resources:
       cpu: "50m"
     limits:
       cpu: "200m"
-      memory: "1Gi"
+      memory: "10Gi"
   backend:
     requests:
       memory: "640Mi"
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -3619,8 +3619,8 @@ defaultResources:
     memory: "10Mi"
     cpu: "10m"
   limits:
-    memory: "1Gi"
-    cpu: "1000m"
+    memory: "200Mi"
+    cpu: "500m"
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -3619,8 +3619,8 @@ defaultResources:
     memory: "10Mi"
     cpu: "10m"
   limits:
-    memory: "1Gi"
-    cpu: "1000m"
+    memory: "200Mi"
+    cpu: "500m"
```
theosanderson: IMO we don't need to throttle the CPU

loculus_values/values.yaml
```diff
@@ -3619,8 +3619,8 @@ defaultResources:
     memory: "10Mi"
     cpu: "10m"
   limits:
-    memory: "1Gi"
-    cpu: "1000m"
+    memory: "200Mi"
```
theosanderson: no strong feelings, but not sure we need this so low

theosanderson (approved):

loculus_values/values.yaml
anna-parker:
```suggestion
    memory: "1Gi"
    cpu: "1000m"
```

PR #922, opened 2026-05-06
fix: remove ncbiReleaseDate from default view in search table for Dengue

### LINKED ISSUE #921: Dates truncated in search table
Some Ncbi release dates are truncated for Dengue

Image

Might just need to broaden the column by few pixels.

corneliusroemer:
fixes #921

Dengue is the only place we show ncbiReleaseDate in search table by default.

The field is mostly redundant with earliest release date, hence not included in all other organisms. Removing for consistency.

@maverbiest as you configured that organism, just checking I'm not missing a reason for including it here but not elsewhere.

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1875,7 +1875,6 @@ organisms:
         tableColumns:
           - sampleCollectionDate
           - earliestReleaseDate
-          - ncbiReleaseDate
           - geoLocCountry
           - authors
           - authorAffiliations


maverbiest (approved):
Thanks for flagging! To be honest I don't remember why this ended up being part of the default columns - likely a holdover from me looking at the loculus values.yaml to learn how to configure things.

So yea perfectly fine to remove

anna-parker:
I am taking the liberty of merging this

PR #926, opened 2026-05-08
Add Andes Virus [Hantavirus] organism

theosanderson-agent:
## Summary

Adds Andes Virus [Hantavirus] as a Pathoplexus organism with three expected genome segments (`L`, `M`, and `S`). The configuration uses the Andes virus NCBI taxon ID `1980456` for ingest, preprocessing, and ENA deposition metadata.

The organism uses Andes virus reference assets staged through GitHub Pages in `pathoplexus/reference_store`:

- base URL: https://pathoplexus.github.io/reference_store/andv/
- `L`: `NC_003468.2`, gene `RdRp`
- `M`: `NC_003467.2`, gene `GPC`
- `S`: `NC_003466.1`, genes `N` and `NSs`

Preprocessing points at the minimal Andes virus Nextclade datasets in `reference_store`, and ingest uses the Andes-specific minimizer for segment identification. The organism image is also rehosted from `reference_store`.

This branch also adds a global multi-field identifier search named `Identifier (INSDC, etc.)` across organisms. It sets `orderInSearchDisplay: -10000` so the field appears near the top of the search UI.

## Validation

- `helm lint pathoplexus_app`
- `git diff --check`
- Ran Helm template checks locally for `demo`, `main`, `production`, and `staging` with the rebased Loculus chart values.
- Verified the configured `https://pathoplexus.github.io/reference_store/andv/...` asset URLs return HTTP 200, including individual Nextclade dataset files.

Preview: https://preview-andes-virus.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2146,7 +2146,7 @@ organisms:
               sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/west-nile/env.fasta]]"
             - name: prM
               sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/west-nile/prM.fasta]]"
-  hanta-andes:
+  andv:
     <<: *defaultOrganismConfig
     schema:
       <<: *schema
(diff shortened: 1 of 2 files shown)


loculus_values/values.yaml
```diff
@@ -2146,6 +2158,118 @@ organisms:
...
+                  genes: [GPC]
+            - name: S
+              references:
+                - name: singleReference
+                  nextclade_dataset_name: andv/S
+                  genes: ["N", NSs]
+    ingest:
+      <<: *ingest
+      configFile:
+        <<: *defaultIngestConfigFile
```
anna-parker, line 2233: Just had a chance to take a look at this - ingest looks great, however I see that a total of `554` segment groups were submitted but only 500 show on the preview. This must mean they are failing preprocessing. 

Logging in as the insdc_ingest_user we have some cases where segments were grouped using the minimizer index but one segment could not align (e.g. M failed to align and S succeeded). We could change the preprocessing conditions to also allow submission of these cases using `alignment_requirement: ANY`, alternatively we could use nextclade align instead of nextclade sort in ingest - as there arent so many andesvirus samples on the INSDC this shouldnt mean a large increase in ingest time

anna-parker: ah you actually use nextclade align for assigning segment in preprocessing - so we would have to use the minimizer instead to use the `alignment_requirement: ANY` option

anna-parker: From looking at the sequences that failed alignment this is typically M sequences - CCHF also has very high diversity in the terminal sections of the M segment - so we might want to tweak the nextclade dataset for this segment to align more.

anna-parker: reducing the min seed cover is typically an easy way to find more alignments in the case of high diversity, I do it for CCHF: https://github.com/nextstrain/nextclade_data/blob/master/data/community/pathoplexus/cchfv/M/pathogen.json#L4

theosanderson: Thanks for these ideas - will make an issue!

theosanderson (approved):

PR #941, opened 2026-05-11
[codex] Update ANDV Nextclade datasets

theosanderson:
## Summary

Update Andes virus Nextclade configuration to use the official Nextclade dataset server paths:

- `nextstrain/orthohantavirus/andv/l`
- `nextstrain/orthohantavirus/andv/m`
- `nextstrain/orthohantavirus/andv/s`

This removes the custom `pathoplexus.github.io/reference_store/andv/nextclade_data` dataset server from the ANDV link-outs and preprocessing config.

## Annotation changes checked

The current official M segment dataset tag (`2026-05-11--18-59-50Z`) adds the requested annotation shape: `GPC` covers the full precursor, and `Gn`/`Gc` cover the split glycoproteins. The Loculus M-segment gene list is updated from `GPC` to `GPC, Gn, Gc`.

L and S keep the expected gene names: `RdRp`, `N`, and `NSs`.

## Preprocessing

Bumped the ANDV preprocessing version from `1` to `2` and pinned all three official datasets to `2026-05-11--18-59-50Z`.

## Validation

- Verified the official Nextclade v3 index contains the new ANDV paths and tag
- Compared old custom and current official GFF annotations for ANDV L/M/S
- `ruby -e 'require "yaml"; YAML.load_file("loculus_values/values.yaml", aliases: true)'`
- `git diff --check`
- `helm template` for `demo`, `main`, `production`, and `staging` overlays

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2172,15 +2172,15 @@ organisms:
         - name: "Nextclade (L)"
           category: "Nextclade"
           maxNumberOfRecommendedEntries: 3000
-          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences:L+rich|fasta]}}&dataset-name=andv/L&dataset-server=https://pathoplexus.github.io/reference_store/andv/nextclade_data"
+          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences:L+rich|fasta]}}&dataset-name=nextstrain/orthohantavirus/andv/l"
         - name: "Nextclade (M)"
           category: "Nextclade"
           maxNumberOfRecommendedEntries: 5000
-          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences:M+rich|fasta]}}&dataset-name=andv/M&dataset-server=https://pathoplexus.github.io/reference_store/andv/nextclade_data"
+          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences:M+rich|fasta]}}&dataset-name=nextstrain/orthohantavirus/andv/m"
         - name: "Nextclade (S)"
           category: "Nextclade"
           maxNumberOfRecommendedEntries: 10000
-          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences:S+rich|fasta]}}&dataset-name=andv/S&dataset-server=https://pathoplexus.github.io/reference_store/andv/nextclade_data"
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -2203,29 +2203,31 @@ organisms:
     preprocessing:
       - <<: *preprocessing
         version:
-          - 1
+          - 2
```
anna-parker, line 2206: when re-processing sequences you need to keep an old prepro version up and running or else all submissions during reprocessing will fail as there is no prepro pod to process them

loculus_values/values.yaml
```diff
@@ -2203,29 +2203,31 @@ organisms:
...
         configFile:
           <<: *preprocessingConfigFile
           create_embl_file: false
           taxon_id: 1980456
           scientific_name: "Andes Virus"
           molecule_type: "genomic RNA"
```
anna-parker, line 2212: here we can switch to 
```
segment_classification_method: "minimizer"
          alignment_requirement: ANY
          minimizer_url: "https://pathoplexus.github.io/reference_store/andv/segment-minimizer.json"
```
probably would be good to double check that you cant submit a valid andes segment and cchf segment together but I think this shoudl work, @corneliusroemer can probably help with this :-)

loculus_values/values.yaml
theosanderson: I haven't finalised this yet but I was actually thinking reprocessing will be very quick so it could be OK

loculus_values/values.yaml
anna-parker: I realized this would allow users to submit an entry with one ANDV segment and another arbitary segment, really we need to merge https://github.com/loculus-project/loculus/pull/6413 to use this

loculus_values/values.yaml
anna-parker: I probaby wouldnt risk it, having one more pod for a couple minutes isnt too much load, and it is our standard rollout procedure

loculus_values/values.yaml
```diff
@@ -2203,29 +2203,31 @@ organisms:
...
+                  nextclade_dataset_tag: "2026-05-11--18-59-50Z"
                   genes: [RdRp]
             - name: M
               references:
                 - name: singleReference
-                  nextclade_dataset_name: andv/M
-                  genes: [GPC]
+                  nextclade_dataset_name: nextstrain/orthohantavirus/andv/m
+                  nextclade_dataset_tag: "2026-05-11--18-59-50Z"
+                  genes: [GPC, Gn, Gc]
```
anna-parker, line 2225: oh wupps I didnt add these new genes to the nextclade dataset update! will make a PR now!

anna-parker: now added in https://github.com/pathoplexus/pathoplexus/pull/1002

PR #944, opened 2026-05-13
fix(config): change hostTaxonId to type: string to match Loculus changes

maverbiest:
To be merged when we roll out https://github.com/loculus-project/loculus/pull/6487.

This PR changes the metadata field `hostTaxonId` from int to string and pins the taxonomy DB to a specific version to match breaking changes in the above PR. In addition to merging this config change, we will need to run an SQL command against the database to convert existing hostTaxonIds to strings in the sequence_entries_preprocessed_data table (see Loculus PR for SQL).

### DIFF
--- a/pathoplexus_app/values.yaml
+++ b/pathoplexus_app/values.yaml
@@ -1,3 +1,3 @@
-loculusVersion: 29e3fca6e5a0ae26e9635d5b909c3a9339ba65d3
+loculusVersion: 0b0b978b68cfd6f863403539cecab0f4450f52ec
 pathoplexusBranch: dummy
 pathoplexusVersion: dummy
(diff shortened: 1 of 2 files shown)


pathoplexus_app/values.yaml
```diff
@@ -1,3 +1,3 @@
-loculusVersion: 29e3fca6e5a0ae26e9635d5b909c3a9339ba65d3
+loculusVersion: f18200ed79b922d29a3903e7afb2b62ada78108b
```
theosanderson, line 1: to revert

maverbiest: Oh yes, thanks! Just for my understanding; when we roll out, we run the action to update the loculus version on main (to a version with the breaking changes), and then point to that updated loculusVersion hash from this branch as well, correct?

theosanderson: The action will make a PR. I'd then probably merge something like this into that PR's branch (whether through a PR or just a push)

anna-parker:
changes now merged

PR #950, opened 2026-05-14
[codex] Update PPX ingest taxon IDs

theosanderson:
## What changed

Updated the Pathoplexus INSDC ingest taxon IDs for:

- dengue: `12637` -> `3052464`
- yellow-fever: `11089` -> `3046277`

## Why

These use the current NCBI species-level parent nodes for the legacy/no-rank virus nodes, matching the intended ingest scope while keeping ENA deposition taxon IDs unchanged.

## Validation

- `python3` YAML parse of `loculus_values/values.yaml`
- `git diff --check`

🚀 Preview: https://preview-taxid2.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1927,7 +1927,7 @@ organisms:
     ingest:
       <<: *ingest
       configFile:
-        taxon_id: 12637
+        taxon_id: 3052464
     enaDeposition:
       DENV-1:
         configFile:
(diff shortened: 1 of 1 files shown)


theosanderson:
@theosanderson-agent add these to the loculus mirror workflow as priority 1

theosanderson-agent:
Done — the mirror workflow lives in `loculus-project/loculus`, so I opened a PR there to add both species-level taxon IDs to the priority-1 schedule (every 2 hours):

- dengue `3052464`
- yellow fever `3046277`

PR: https://github.com/loculus-project/loculus/pull/6445

Note: the legacy dengue id `12637` is still in `datasets-mirror-priority-2.yml` — left it alone, but it can be dropped once nothing depends on it.

PR #964, opened 2026-05-20
Add outbreak annotation to Bundibugyo dataset, reprocess with new Nextclade dataset

corneliusroemer:
- **Reprocess bdbv with new dataset version**
- **No need for exact dataset version in linkout to nextclade**
- **Lift outbreak annotation and display in table**

image

🚀 Preview: https://preview-bdbv-outbreak.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1,3 +1,6 @@
+# REVERT DONT MERGE
+createTestAccounts: true
+# REVERT ABOVE
 preprocessingTimeout: 600
 getSubmissionListLimitSeconds: 3600
 robotsNoindexHeader: true
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1785,7 +1785,7 @@ organisms:
       linkOuts:
         - name: "Nextclade"
           maxNumberOfRecommendedEntries: 3000 # ebola is ~18.5kb
-          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences+rich|fasta]}}&dataset-name=nextstrain/orthoebolavirus/bdbv&dataset-tag=2026-05-15--16-16-45Z"
+          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences+rich|fasta]}}&dataset-name=nextstrain/orthoebolavirus/bdbv"
         - name: "Sealion"
           maxNumberOfRecommendedEntries: 3000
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
I havent looked at this but I trust you ran the regression test to compare the changes this will introduce :-) (would be cool if you could maybe summarize those findings)

loculus_values/values.yaml
```diff
@@ -1785,26 +1785,65 @@ organisms:
       linkOuts:
         - name: "Nextclade"
           maxNumberOfRecommendedEntries: 3000 # ebola is ~18.5kb
-          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences+rich|fasta]}}&dataset-name=nextstrain/orthoebolavirus/bdbv&dataset-tag=2026-05-15--16-16-45Z"
+          url: "https://clades.nextstrain.org/?input-fasta={{[unalignedNucleotideSequences+rich|fasta]}}&dataset-name=nextstrain/orthoebolavirus/bdbv"
```
anna-parker, line 1788: ah wait maybe we add the `2026-05-18--20-09-34Z` tag incase we update the dataset in the future?

corneliusroemer: I don't see a need to pin the linkout dataset version - one will want to analyze with the most up to date dataset in those cases no?

corneliusroemer:
> I havent looked at this but I trust you ran the regression test to compare the changes this will introduce :-) (would be cool if you could maybe summarize those findings)

I haven't - we'd see during standard rollout procedure via staging anyways?

loculus_values/values.yaml
anna-parker: I think we should linkout to the same dataset we processed the data with (IMO), but maybe thats an opinion

anna-parker: tagging @theosanderson if he has time to give a third opinion

corneliusroemer: We're not doing that anywhere but for bdbv, it's one more thing we'd need to update and check etc. Simpler to just leave it out and point at the latest released dataset

theosanderson: I don't have a strong view - my agent added the tag and I didn't bother to change it but I probably agree a little with cornelius

corneliusroemer: The fact that we use Nextclade in prepro is unrelated to the linkouts in my view, so we shouldn't burden ourselves with maintenance and send people to outdated datasets.

PR #965, opened 2026-05-20
Show YFV clade in search table and move Dengue/CCHF lineage/clade to left to canonical position

corneliusroemer:
- **Show YFV clade in default search table view**
- **Move CCHF segment S lineage to canonical lineage/clade position on left side in search table**
- **Move dengue lineage left in search table**

Fixing YFV table columns to include `clade`. It was `lineage` but that doesn't exist

Now all lineage/clade/outbreak columns across all organisms are immediately to the right of the accession column for uniform appearance.

image

image

image

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2400,6 +2400,7 @@ organisms:
             inputs: {input: nextclade.clade}
           order: 5
           orderOnDetailsPage: 740
+          columnWidth: 60
       website:
         <<: *website
         tableColumns:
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -1910,6 +1910,8 @@ organisms:
           includeInDownloadsByDefault: true
           preprocessing:
             inputs: {input: nextclade.clade}
+          order: 5
```
theosanderson, line 1913:
```suggestion
          order: 5
          columnWidth: 80
```

theosanderson (approved):

PR #970, opened 2026-05-21
Rename Tools nav item to Resources

theosanderson:
Change the top navigation menu label from "Tools" to "Resources" and update the icon to tabler:presentation, and update text

image

🚀 Preview: https://preview-resource.pathoplexus.org

### DIFF
--- a/monorepo/website/src/components/ExternalTools/NoToolsAvailable.astro
+++ b/monorepo/website/src/components/ExternalTools/NoToolsAvailable.astro
@@ -3,6 +3,6 @@
 
 <div class="text-center py-12">
     <p class="mt-2 text-gray-500">
-        There are currently no tools available.
+        There are currently no resources available.
     </p>
 </div>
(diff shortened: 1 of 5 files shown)


chaoran-chen (approved):
Thanks Theo! I'm happy with the new wording. The icon is less intuitive to me but no strong opinion. I think it's good that it's not changing the URL path, avoiding the need to introduce redirects (or break links).

theosanderson:
@theosanderson-agent redo the path to resources and add a redirect from current path

theosanderson:
Thanks Chaoran. I felt like the wrench icon was relatively unintuitive, I'd have expected it to change settings or something, so that's why this. Definitely open to it being changed in future, I don't think it's perfect.

theosanderson-agent:
Done in 343cf8d:
- Renamed the page route from `/[organism]/external-tools` to `/[organism]/external-resources`
- Updated the Resources nav item to point at the new path
- The old `/[organism]/external-tools` path now serves a 301 redirect to the new location

PR #974, opened 2026-05-22
add search function for docs & about!

emmahodcroft:
Watch out everyone - I'm out here being a dev thanks to Claude! 👊 💥 pow, pow!

It's quite hard to find things in our docs - the menu on the left is quite overwhelming to read through. If you know exactly what you want, it's ok, but for a newcomer I imagine it's quite intimidating.

So let's fix it! Claude and I did it together. [Here's](https://github.com/loculus-project/loculus/pull/6496/changes) his take on this PR. I tested it locally and it seems to work 😁 

That'll do pig. That'll do.

🚀 Preview: https://preview-try-docs-search.pathoplexus.org


monorepo/website/src/pages/search-index.json.ts
```diff
@@ -0,0 +1,11 @@
+import { buildSearchIndex } from '../utils/buildSearchIndex';
+
+export async function GET() {
+	const indexJson = await buildSearchIndex();
+
+	return new Response(indexJson, {
+		headers: {
```
theosanderson: We should probably at least add some headers that allow caching this for a while to reduce how often this gets called (both to reduce amount transferred to user and to reduce load on us)

monorepo/website/src/utils/buildSearchIndex.ts
```diff
@@ -0,0 +1,143 @@
...
+	// /path/to/src/pages/docs/how-to/index.mdx -> /docs/how-to
+	const parts = filePath.split('/pages/')[1];
+	const urlPath = parts
+		.replace(/\.mdx$/, '')
+		.replace(/\/index$/, '');
+
+	return `/${urlPath}`;
+};
+
+const getTitleFromMdx = (content: string): string => {
```
theosanderson: It would be nicer to properly parse the yaml I think

monorepo/website/src/utils/buildSearchIndex.ts
```diff
@@ -0,0 +1,143 @@
...
+
+	for (const { path, section } of allFiles) {
+		try {
+			const content = readFileSync(path, 'utf-8');
+			const title = getTitleFromMdx(content);
+			const textContent = extractTextFromMdx(content);
+			const url = parseUrlFromPath(path);
+
+			// Skip index pages as they'll be redundant
+			if (!url.endsWith('index')) {
```
theosanderson: instead we should probably only include mdx files

monorepo/website/src/pages/search-index.json.ts
corneliusroemer: hash of the index seems like the obvious cache key

monorepo/website/src/utils/buildSearchIndex.ts
theosanderson: hmm maybe we already do? but if so we wouldn't need this?

corneliusroemer (commented):
Need to add minisearch to deps that's why image build fails - do we allow adding new packages in the overlay? Might be tricky.

Overriding whole package.json would be messy. So for now would have to add the package on Loculus side

emmahodcroft:
No this is me not knowing about dev things. Theo says we should add `minisearch` (and now it would like to add `yaml` to address some of Theo's comments) to Loculus

monorepo/website/src/pages/search-index.json.ts
theosanderson: I'm saying a cache that's just like 1 hr TTL so that it doesn't actually try to rebuild the index (for that user) in that hour

theosanderson: (some time we could build the index as part of the build step but this seems simple for now)

theosanderson:
@theosanderson-agent - try to address all the points above. you can push straight to Emma's branch. You need to make a PR to loculus to add the necessary dependencies, e.g. minisearch, yaml. Then you can update the loculusVersion in this PR to point to the commit hash from the Loculus PR and check CI passes here.

monorepo/website/src/pages/search-index.json.ts
theosanderson-agent: Moved index generation to build time (via `import.meta.glob`) in 9fbe198, so the JSON is precomputed when the server boots and the route just hands it back. With the work out of the request path the ETag / `Cache-Control` headers in the earlier round (c609ad2) became redundant and have been removed.

monorepo/website/src/utils/buildSearchIndex.ts
theosanderson-agent: Frontmatter is now parsed with the `yaml` library in c609ad2 — the regex-based title extraction is gone.

monorepo/website/src/utils/buildSearchIndex.ts
theosanderson-agent: Right, the `if (!url.endsWith("index"))` check was dead — `parseUrlFromPath` already strips trailing `/index`. Removed in c609ad2; index pages now get included.

theosanderson-agent:
Took a pass at the review here:

- **Caching / hash cache key** — Moved the index build to *build time* in 9fbe198 using `import.meta.glob` with `?raw`; Vite bundles the MDX content into the server bundle and the payload is `JSON.stringify`-ed once when the Node server boots. The route is now a one-line `new Response(...)`. With the work out of the request path the ETag / `Cache-Control` / 304 plumbing from the first round (c609ad2) became redundant and is gone.
- **Properly parse the yaml** — Frontmatter is now parsed with the `yaml` library (c609ad2).
- **Only mdx files / redundant `index` skip** — Confirmed `parseUrlFromPath` already strips trailing `/index`, so the `if (!url.endsWith("index"))` guard was dead code. Removed in c609ad2; index pages now get included.

Other things along the way:

- Loculus PR loculus-project/loculus#6496 (deps) needed a `package-lock.json` regen — the first attempt regenerated the whole lockfile and accidentally marked vite's nested platform `@esbuild/*` binaries as `dev: true`, which broke `npm ci --omit=dev` in the website Docker build with an esbuild postinstall mismatch. Reverted to a surgical 8-line diff that only adds the `minisearch` package entry (46b6cf6).
- `loculusVersion` bumped to that commit on this PR.
- Small cleanups in `DocsSearch.tsx`: fixed the `SearchResult` cast, typed the fetched response, swapped the clear button for the wrapped `Button` from `common/`, and removed lint-flagged `console.error` / floating promise issues.

Build is green on both PRs.

monorepo/website/src/pages/search-index.json.ts
theosanderson: (still investigating this)

theosanderson: now built on build/startup (or maybe it always was, and using etag)

pathoplexus_app/values.yaml
```diff
@@ -1,3 +1,3 @@
-loculusVersion: 30f3557c5c7066a1e1bfdb7d586f4cca3f4c64c9
+loculusVersion: 3117e2f5e27a46f5197eda78fda2f19f6ef12285
```
theosanderson, line 1: revert

theosanderson:
merged in loculus

PR #987, opened 2026-05-29
feat: increase silo import pollIntervalSeconds to 2min

anna-parker:
increase poll interval to 2min for all previews and staging (and prod for now - lower after rollout/reprocessing complete)

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -70,7 +70,7 @@ sequenceFlagging:
 siloImport:
   siloTimeoutSeconds: 3600
   hardRefreshIntervalSeconds: 100_000 # ~27 hours
-  pollIntervalSeconds: 30
+  pollIntervalSeconds: 120
 pipelineVersionUpgradeCheckIntervalSeconds: 300
 zstdCompressionLevel: 20
 lineageSystemDefinitions:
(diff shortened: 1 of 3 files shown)


loculus_values/values.yaml
```diff
@@ -70,7 +70,7 @@ sequenceFlagging:
 siloImport:
   siloTimeoutSeconds: 3600
   hardRefreshIntervalSeconds: 100_000 # ~27 hours
-  pollIntervalSeconds: 30
+  pollIntervalSeconds: 120
```
theosanderson, line 73: As implemented this will affect prod as far as I can see which would be problematic

corneliusroemer:
Plan is to lower for prod again after the rollout

loculus_values/values.yaml
anna-parker: we were thinking of just increasing during the reprocessing/release and then decreasing but I can add this specifically in the config!

theosanderson:
ok, but you could also just add a prod setting?

theosanderson:
oh wait you are intentionally lowering(raising) it for prod?

theosanderson:
ok

theosanderson:
We need to get in the logic from https://github.com/loculus-project/loculus/pull/5111 and I think that would resolve all this stuff (but not now, ofc)

PR #988, opened 2026-05-29
Use separate 100s ranges for each organism's prepro version for db efficiency

corneliusroemer:

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1676,7 +1676,7 @@ organisms:
         replicas: 1
         version:
           - 27
-          - 28
+          - 100
         configFile:
           <<: *preprocessingConfigFile
           segments:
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
theosanderson, line 1: This will be hard to roll-back from - pls can we discuss in a meeting first and spend some time figuring out if there's an alternative

theosanderson: [or give time for me to spend that time I mean]

fhennig:
Context from Slack

> If we don't reuse pipeline versions for different organisms then the queries to upgrade pipeline version and to get data for prepro become faster - as the join/where becomes more selective.
As prepro version is just an arbitrary number this should have no downsides for us other than a little cognitive effort for maintainin values.yaml.

([by Cornelius](https://loculus.slack.com/archives/C0A232FGH61/p1765787390269949))

PR #994, opened 2026-06-01
test reducing zstd to 15 (also slow silo import to separate out)

corneliusroemer:
- **test if rarely calling get released data (20min) reduces backend cpu measurably**
- **Reduce zstd level to 15**


anna-parker:
chnages are now merged :-)

PR #1002, opened 2026-06-01
feat(andv): add more genes

anna-parker:
Genes added to cornelius' repo using script: https://github.com/corneliusroemer/seqs/tree/main/artefacts/orthohantavirus and now added here.

This PR will bump staging pipeline version from 3 to 4 and prod from 2 to 4 - thus it is safe to merge now and include in the release today

image

- [x] confirm no gene/reference has changed with diff
I diffed every gene and reference to ensure there were no changes!

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2350,7 +2350,7 @@ organisms:
                   genes: ["N", NSs]
       - <<: *preprocessing
         version:
-          - 3
+          - 4
         configFile:
           <<: *preprocessingConfigFile
           create_embl_file: false
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -2371,7 +2371,7 @@ organisms:
               references:
                 - name: singleReference
                   nextclade_dataset_name: nextstrain/orthohantavirus/andv/m
-                  genes: [GPC]
+                  genes: [GPC, Gn, Gc]
             - name: S
               references:
                 - name: singleReference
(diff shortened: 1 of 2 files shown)


loculus_values/values.yaml
```diff
@@ -2397,29 +2397,33 @@ organisms:
...
           - name: singleReference
-            sequence: "[[URL:https://pathoplexus.github.io/reference_store/andv/NC_003467.2.fasta]]"
+            sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/orthohantavirus/andv/m/reference.fasta]]"
             insdcAccessionFull: NC_003467.2
             genes:
               - name: GPC
-                sequence: "[[URL:https://pathoplexus.github.io/reference_store/andv/genes/GPC.fasta]]"
+                sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/orthohantavirus/andv/m/GPC.fasta]]"
+              - name: Gn
+                sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/orthohantavirus/andv/m/Gn.fasta]]"
```
theosanderson, line 2414: We should probably keep all the files for an organism in one place? (IMO we should ideally be aiming for a place owned by pathoplexus.)

theosanderson: Ah maybe it all is in one place now!

theosanderson (approved):

PR #1014, opened 2026-06-05
feat: apply guaranteed QoS for all required/user-facing pods and reduce requested silo/lapis mem for ebola-bdbv and andv

### LINKED ISSUE #920: Ensure all production workflows have memory `requests` = `limits`
Per conversation with @anna-parker 

For context this means that we would like to ensure all prod pod have guaranteed QoS: https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#guaranteed. 

Potentially ingest can be excluded.

anna-parker:
resolves https://github.com/pathoplexus/pathoplexus/issues/920

andv and ebola-bdbv have only very few sequences, we should provision them similarly to yellow-fever and the other ebolas.

Ingest for small organism requests roughly 400 times more mem and cpu than is required (this reduces mem and cpu for ingest by a factor of 10 for small organisms), silo of andv and ebola-bdbv request 25times more than is required (this reduces silo and lapis by a factor of 2).

looking at https://grafana.pathoplexus.org/d/cfl58hc2l83cwb/requested-resources-vs-used-resources?orgId=1&editPanel=2&from=now-1h&to=now apart from ingest the requested memory was actually quite well aligned with the used memory, sth like a factor of 10more on average

### DIFF
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -79,6 +79,7 @@ enaDeposition:
   enaUniqueSuffix: Pathoplexus
   enaIsBroker: True
 resources:
+# For backend and other required pods, set memory requests = memory limits to ensure guaranteed QoS
   website:
     requests:
       memory: "2Gi"
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -179,6 +179,13 @@ resources:
           cpu: "250m"
         limits:
           memory: "1Gi"
+      ingest:
+        requests:
+          memory: "200Mi"
+          cpu: "50m"
+        limits:
+          cpu: "200m"
+          memory: "10Gi"
     yellow-fever:
       <<: *smallOrganismQueryEngineResources
     ebola-sudan:
--- a/loculus_values/environment_specific_values/production.yaml
+++ b/loculus_values/environment_specific_values/production.yaml
@@ -199,6 +199,32 @@ resources:
           cpu: "250m"
         limits:
           memory: "2Gi"
+    ebola-bdbv:
+      silo:
+        requests:
+          memory: "1Gi"
+          cpu: "250m"
+        limits:
+          memory: "1Gi"
+      lapis:
+        requests:
+          memory: "220Mi"
+          cpu: "250m"
+        limits:
+          memory: "2Gi"
+    andv:
+      silo:
+        requests:
+          memory: "1Gi"
+          cpu: "250m"
+        limits:
+          memory: "1Gi"
+      lapis:
+        requests:
+          memory: "220Mi"
+          cpu: "250m"
+        limits:
+          memory: "2Gi"
 defaultResources:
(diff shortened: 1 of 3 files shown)


loculus_values/environment_specific_values/production.yaml
```diff
@@ -199,6 +199,32 @@ resources:
...
         limits:
           memory: "2Gi"
+    ebola-bdbv:
+      silo:
+        requests:
+          memory: "1Gi"
+          cpu: "250m"
+        limits:
+          memory: "1Gi"
+      lapis:
```
theosanderson: per our previous discussions we wanted to ensure that memory always matches limits in production to avoid eviction (I'm not averse to lower limits)

anna-parker: ah we actually use this above as well - so I will change there too

anna-parker: (I just copied the config)

loculus_values/values.yaml
```diff
@@ -3919,7 +3920,7 @@ resources:
       memory: "10Gi"
   backend:
     requests:
-      memory: "640Mi"
+      memory: "3Gi"
```
theosanderson: how come we're doing this even for previews?

anna-parker: ah true I guess we only need guaranteed QoS for prod - I will remove this change :-)

theosanderson:
@codex review

loculus_values/environment_specific_values/production.yaml
```diff
@@ -172,39 +173,34 @@ resources:
...
+      ingest:
         requests:
-          memory: "220Mi"
-          cpu: "250m"
+          memory: "100Mi"
+          cpu: "20m"
         limits:
-          memory: "2Gi"
+          cpu: "200m"
+          memory: "1Gi"
```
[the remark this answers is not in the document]
anna-parker: wait what? have these never been applied?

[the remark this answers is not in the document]
theosanderson: If I'm reading correctly we have never attempted to specify per-organism ingest resources until this PR (and the model is right that we are not set up to do so)

loculus_values/environment_specific_values/production.yaml
```diff
@@ -73,18 +73,19 @@ seqSets:
...
 resources:
+# For backend and other required pods, set memory requests = memory limits to ensure guaranteed QoS
   website:
     requests:
-      memory: "1Gi"
+      memory: "2Gi"
       cpu: "1000m"
     limits:
-      memory: "5Gi"
+      memory: "2Gi"
```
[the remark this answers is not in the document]
theosanderson: I asked Claude about this. It gets complicated. The set up we have in this PR will protect from eviction (by the kubelet). What it won't do is to protect from OOM-killing if memory is truly exhausted on the node (by the kernel). But that killing is independent of pod-priority class anyway so there is little we can do there and I think that's ok.

PR #1025, opened 2026-06-08
feat(config): preprocessing no longer uses `hostNameScientific` and `hostTaxonId` as fallback values - remove them

### LINKED ISSUE #1024: [BUG] Host metadata field warnings are confusing/invalid
**Describe the bug**
Reported by Anderson

When submitting dengue sequences to PPX with the metadata field `hostNameScientific` set to `Homo Sapiens` preprocessing returns an error:
```
Host species: Metadata field `hostTaxonld` is required. Please provide input metadata field(s): `host`, `hostNameScientific`
```
This is an outdated error message, now we require users submit `host` and we validate that field and produce the fields `hostTaxonld` and `hostNameScientific`.

anna-parker:
…hostTaxonId` as fallback values and requires `host`, remove these values from the config to prevent invalid and confusing error messages

resolves: https://github.com/pathoplexus/pathoplexus/issues/1024

I created an account on the preview and validated that the message is now clear:
image
I then confirmed that using host results in valid output:
image

🚀 Preview: https://preview-host.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1286,7 +1286,6 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: scientific_name_from_id
           inputs:
             hostTaxonId: processed.hostTaxonId
-            hostNameScientific: hostNameScientific
           args:
             taxonomy_service_url: &taxonomy_service_url "http://loculus-taxonomy-service:5000"
       - name: hostNameCommon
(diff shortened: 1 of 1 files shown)


maverbiest (approved):
Looks good! Thanks a lot for the fix (and to Anderson for reporting)

PR #1027, opened 2026-06-09
Hint about using all data in outbreak settings

theosanderson:

### DIFF
--- a/monorepo/website/src/pages/about/terms-of-use/data-use-terms.mdx
+++ b/monorepo/website/src/pages/about/terms-of-use/data-use-terms.mdx
@@ -145,7 +145,7 @@ The focus of Pathoplexus is on open availability of data. If Users share Restric
 
 <OutbreakHintBox>
     Treating your entire dataset as a "Background Set" is generally **not appropriate in an outbreak
-    setting**. During an outbreak, the majority of the most important and timely data is typically
+    setting**. During an outbreak, the majority of the most important and timely data can be
     Restricted-Use and has been contributed by only a small number of submitting groups. In that
     situation, using all data without contacting and involving those submitters would not be ethical —
     such data should be treated as part of your [Focal Set](#423-deciding-if-restricted-use-data-is-part-of-a-focal-or-background-set)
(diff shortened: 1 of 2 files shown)


corneliusroemer:
https://preview-outbreak.pathoplexus.org/about/terms-of-use/data-use-terms

Google Chrome 2026-06-09 15 17 02

emmahodcroft:
LGTM my only question would be a slight qualm on "is typically Restricted-Use" - maybe 'can be Restricted-Use' ? In theory there could be an outbreak where a lot of data is released openly?

But I might be overthinking this.

emmahodcroft:
Oh want wait - should this be in 'using all data'? I don't think so

Should probably be wherever we discuss Focal & Restricted use (which I think is one section up)

monorepo/website/src/pages/about/terms-of-use/data-use-terms.mdx
```diff
@@ -141,6 +143,15 @@ The focus of Pathoplexus is on open availability of data. If Users share Restric
...
 
+<OutbreakHintBox>
+    Treating your entire dataset as a "Background Set" is generally **not appropriate in an outbreak
+    setting**. During an outbreak, the majority of the most important and timely data can be
+    Restricted-Use and has been contributed by only a small number of submitting groups. In that
+    situation, using all data without contacting and involving those submitters would not be ethical —
+    such data should be treated as part of your [Focal Set](#423-deciding-if-restricted-use-data-is-part-of-a-focal-or-background-set)
+    and used under the [Restricted Data Use](#42-restricted-data-use) requirements.
+</OutbreakHintBox>
+
```
anna-parker, line 155: Hmm... I guess this is only if the outbreak data is restricted use, claude suggested:
```
Treating your entire dataset as a "Background Set" may **not be appropriate in an outbreak
    setting when key outbreak data are Restricted-Use**. During an outbreak, the most important and
    timely data may come from a small number of submitting groups and may still be under Restricted-Use
    protections. In that situation, using those data as background without contacting and involving the
    submitters would not be appropriate — such data should instead be treated as part of your
    [Focal Set](#423-deciding-if-restricted-use-data-is-part-of-a-focal-or-background-set)
    and used under the [Restricted Data Use](#42-restricted-data-use) requirements.
```

emmahodcroft: I like this slight modification to the wording

anna-parker (approved):
Thanks! Always fun to have claude apply claude's suggestions

PR #1033, opened 2026-06-10
Use fixed gray border on cards instead of theme base color

theosanderson:
https://pathoplexus.org/browse/select-organism lost its borders (almost) in the Daisy upgrade. This restores them.

__________

The organism selector cards (`/browse/select-organism`, `/submission/select-organism`) and the homepage cards render with a near-invisible border in production. The cards use `border border-base-200/70 bg-base-100`.

The DaisyUI 5 / Tailwind v4 migration (#997) changed the light theme's `base-200` from `oklch(93% 0 0)` (≈`#e5e6e6`) to `oklch(98% 0 0)` (≈`#f8f8f8`). Since the card background (`base-100`) is white, a 98%-lightness border at 70% opacity is barely distinguishable from the background — so the cards lost the defined border they had before the migration.

## Fix

Replace the theme `base` tokens with fixed Tailwind palette values that keep the border lightness stable regardless of DaisyUI theme changes:

- `border-base-200/70` → `border-gray-200/70` — Tailwind `gray-200` is `oklch(92.8%)` (≈`#e5e7eb`), matching the earlier `base-200` lightness of `oklch(93%)`. The `/70` opacity is preserved.
- `bg-base-100` → `bg-white` — `base-100` was white both before and after the migration; this just removes the remaining adaptive token.

Applied to all six occurrences across the two files that used these tokens:
- `monorepo/website/src/components/OrganismSelector/CardsToSelectOrganism.astro`
- `monorepo/website/src/pages/index.astro`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

🚀 Preview: https://preview-organism-card-border-gray.pathoplexus.org


tombch (approved):
Had a look at the preview, looks good to me!

PR #1047, opened 2026-06-19
Update docs for host metadata field to address waste water

maverbiest:
This PR extends the guidance for the `host` metadata field to also specify what to do when submitting data derived from waste water (as well as cases where the host is unidentified in general).

I also changed `Samples pooled` to `pooling specimens`, since that appears to be the current name attached to OBI:0600016: http://purl.obolibrary.org/obo/OBI_0600016

I also considered adding a small section on waste water submissions to `monorepo/website/src/pages/docs/how-to/upload-sequences.mdx` but it didn't really seem to fit super well. I'm also not sure if submissions of waste water derived consensus sequences data is something we want promote at the moment. For these reasons I decided to just extend the guidance on the metadata field, but please let me know if I should extend the `how-to` docs as well!

🚀 Preview: https://preview-waste-water-host.pathoplexus.org


loculus_values/values.yaml
```diff
@@ -1571,7 +1571,7 @@ defaultOrganismConfig: &defaultOrganismConfig
...
+        guidance: "You can provide either the scientific name of the organism (e.g., `Aedes aegypti`) or its NCBI taxon ID (e.g., `7159`) in this field. We will validate your input and use it to populate the `hostNameCommon` and `hostNameScientific` fields. If the host is not known, please fill in `unidentified`. You should also fill in `unidentified` when submitting data derived from waste water (in addition to setting the `environmentalMaterial` field to `waste water`, see [ENVO:00002001]). When uploading data from a pooled sample, we recommend using a higher-level taxon that includes all host organisms in the sample pool. For example, you can use `Culicidae` when uploading data for a pooled mosquito sample (in addition to setting the `specimenProcessing` field to `pooling specimens`, see [OBI:0600016])."
```
theosanderson:
```suggestion
        guidance: "You can provide either the scientific name of the organism (e.g., `Aedes aegypti`) or its NCBI taxon ID (e.g., `7159`) in this field. We will validate your input and use it to populate the `hostNameCommon` and `hostNameScientific` fields. If the host is not known, please enter `unidentified`. You should also enter `unidentified` when submitting data derived from waste water (in addition to setting the `environmentalMaterial` field to `waste water`, see [ENVO:00002001]). When uploading data from a pooled sample, we recommend using a higher-level taxon that includes all host organisms in the sample pool. For example, you can use `Culicidae` when uploading data for a pooled mosquito sample (in addition to setting the `specimenProcessing` field to `pooling specimens`, see [OBI:0600016])."
```

anna-parker: potentially we can also add a FAQ for submitting wastewater where it mentions these fields :-)

theosanderson: I guess my take would be that I'm not certain we need to specify particularly behaviour for wastewater here

anna-parker: Where would you think is a better place for specifying this? in the docs or FAQ?

theosanderson: I think we need to tell people how to specify information on hosts, including when they don't know the host species. That makes sense here.

I think we need to tell people how to mark something as a wastewater sample - in the environment metadata sections I guess.

Personally I don't think we need specific advice on what host data to supply for wastewater. I think we can leave this up to submitters. E.g. maybe for a city's wastewater system one might feel one doesn't know the host, whereas for an aeroplane toilet maybe one is confident its human. I think we can just leave this to submitter judgement.

maverbiest:
@theosanderson @anna-parker I now removed mention of wastewater from the `host` guidance, added a section above the metadata fields table, and added a FAQ entry on wastewater. Let me know what you think!

loculus_values/values.yaml
maverbiest: I added a FAQ entry for wastewater

monorepo/website/src/pages/about/faq.mdx
```diff
@@ -113,6 +113,15 @@ At submission, you can choose whether you’d like your data to be [protected fo
 Once the data is open, it also appears on INSDC-member databases.
 </details>
 
+<details>
+<summary> <span class="faq">Can I contribute my wastewater sequencing data to Pathoplexus?</span> </summary>
+Pathoplexus currently does not accept uploads of raw sequencing reads. However, you may submit consensus sequences derived from wastewater sequencing data.
+
+If you do, please enter the appropriate standardized term in the `environmentalMaterial` metadata field when submitting your data (e.g., waste water [ENVO:00002001]).
+For some organisms on Pathoplexus, the `host` metadata field is required.
+Should the host organism for your wastewater sample be unknown, you may enter 'unidentified' in the `host` field.
```
anna-parker: I think this is typically `unknown` - so maybe this could be rephrased

anna-parker: we should add that we recommend submitting raw reads to the INSDC for waste water

monorepo/website/src/pages/about/faq.mdx
```diff
@@ -113,6 +113,15 @@ At submission, you can choose whether you’d like your data to be [protected fo
 Once the data is open, it also appears on INSDC-member databases.
 </details>
 
+<details>
+<summary> <span class="faq">Can I contribute my wastewater sequencing data to Pathoplexus?</span> </summary>
+Pathoplexus currently does not accept uploads of raw sequencing reads. However, you may submit consensus sequences derived from wastewater sequencing data.
+
+If you do, please enter the appropriate standardized term in the `environmentalMaterial` metadata field when submitting your data (e.g., waste water [ENVO:00002001]).
```
anna-parker: I think we should probably ask them to submit specifically `environmentalMaterial:waste water` to allow better filtering

maverbiest: I rephrased it a bit, is it better now? Or did literally mean to say people should enter `environmentalMaterial:waste water` with that formatting specifically? We don't ask people to do that for other metadata fields where we point to an ontology

monorepo/website/src/pages/about/faq.mdx
maverbiest: Alright I rephrased and mention INSDC for raw reads

monorepo/website/src/pages/about/faq.mdx
```diff
@@ -113,6 +113,15 @@ At submission, you can choose whether you’d like your data to be [protected fo
 Once the data is open, it also appears on INSDC-member databases.
 </details>
 
+<details>
+<summary> <span class="faq">Can I contribute my wastewater sequencing data to Pathoplexus?</span> </summary>
+Pathoplexus currently does not accept uploads of raw sequencing reads. We recommend submitting your raw reads to one of the [INSDC](https://www.insdc.org/) databases. However, you may submit consensus sequences derived from wastewater sequencing data to Pathoplexus.
+
+If you do, please enter 'waste water [ENVO:00002001]' in the `environmentalMaterial` metadata field when submitting your data.
+For some organisms on Pathoplexus, the `host` metadata field is required.
+Since the host organism for wastewater samples is likely to be unknown, you may enter 'unidentified' in the `host` field.
```
anna-parker:
```suggestion
If you do, please enter `waste water` (which corresponds to [ENVO:00002001]) in the `environmentalMaterial` metadata field when submitting your data.

For some organisms on Pathoplexus, the `host` metadata field is required. Since wastewater samples do not have a clearly defined host organism, please enter `undefined` in the `host` field for organisms where `host` is required.
```

anna-parker (approved):
looking great!

PR #1070, opened 2026-07-07
Reprocess mpox with new mpox dataset, support new outbreakLineage field for sublineage search

corneliusroemer:
Mpox now has lineages under a second outbreak `sh2023` (clade Ib) so we can no longer use hierarchical search on lineages as lineages are ambiguous (A.1 could mean `sh2023/A.1` or `sh2023/A.1` and in particular aliases will be different.

Hierarchical search now needs to use the new `outbreakLineage` field - this is part of the new mpox dataset, so we reprocess with that.

We now show outbreak between clade and lineage in search table by default to be unambiguous (there's also outbreak `sh2024` that doesn't have a lineage system at this point).

After reprocessing complete, in next rollout, lineage should no longer be used for sublineage search as it's ambiguous (one name refers to 2 different things, `A.1` includes both clade Ib and IIb outbreak sequences).

Also updates AGENT.md to not ask to run ./sync.sh when it's not necessary (agents would always run it even though I didn't need it)

### Testing

Google Chrome 2026-07-17 15 58 57

image

- Same number of mpox sequences as on main
- Hierarchical search works with new outbreakLineage field

🚀 Preview: https://preview-mpox-ib-lineages.pathoplexus.org


loculus_values/values.yaml
```diff
@@ -76,6 +76,10 @@ zstdCompressionLevel: 15
 lineageSystemDefinitions:
   mpox:
```
corneliusroemer, line 78: `mpox` system should be removed after reprocessing is complete on next rollout after this. For now it's kept so lineage stays searchable until reprocessing is done.

loculus_values/values.yaml
```diff
@@ -2563,7 +2567,7 @@ organisms:
       metadataAdd:
         - name: clade
           displayName: Clade
-          definition: "Clade label assigned by Nextclade, see https://nextstrain.org/fetch/data.clades.nextstrain.org/v3/nextstrain/mpox/all-clades/2025-12-16--20-07-31Z/tree.json for a view of the Nextclade dataset."
+          definition: "Clade label assigned by Nextclade"
```
anna-parker: @tombch and I intentionally added the links to nextclade - I think it would be great to keep them just with the updated links

loculus_values/values.yaml
```diff
@@ -2592,20 +2596,40 @@ organisms:
...
+          definition: Lineage assigned by Nextclade
           isSequenceFilter: true
           header: "Clade & Lineage"
           includeInDownloadsByDefault: true
           noInput: true
           generateIndex: true
           autocomplete: true
+          # should be set to false on next update (after reprocessing complete)
+          # as lineage itself is no longer unambiguous, outbreak is required
+          # otherwise filtering by A.1 includes sequences from 2 outbreaks
```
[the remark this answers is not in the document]
anna-parker: Im not sure I understand your comment, is claude correct?

loculus_values/values.yaml
```diff
@@ -2592,20 +2596,40 @@ organisms:
...
           initiallyVisible: true
           preprocessing:
             inputs: {input: nextclade.customNodeAttributes.lineage}
           columnWidth: 64
           order: 7
           orderOnDetailsPage: 740
           lineageSystem: mpox
```
anna-parker, line 2616: shouldnt we just remove the `lineageSystem: mpox` here?

loculus_values/values.yaml
corneliusroemer, line 78:
```suggestion
  # TODO: remove mpox lineage system after preprocessing with mpoxOutbreakLineage is complete
  mpox:
```

loculus_values/values.yaml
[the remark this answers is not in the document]
corneliusroemer: Yeah not wrong, I'll update before merging - not now because it would wipe the preview

loculus_values/values.yaml
corneliusroemer: Do these show anywhere user visible? Otherwise not sure it's worth maintaining

corneliusroemer: I see here, hmm, https://pathoplexus.org/docs/concepts/metadataformat?organism=mpox&fieldType=generatedFields 
Google Chrome 2026-07-20 10 06 17

loculus_values/values.yaml
corneliusroemer: Then while rollout happens there's no lineage search but sure could do that already

.gitignore
```diff
@@ -14,3 +14,5 @@ monorepo/**/GEMINI.md
 # Exception: Allow AGENTS.md symlinks in docs
 !monorepo/website/src/pages/docs/CLAUDE.md
 !monorepo/website/src/pages/docs/GEMINI.md
+
+/tmp/
```
corneliusroemer:
```suggestion
/tmp/

```

loculus_values/values.yaml
anna-parker: could you please add them back? This was requested in the past

loculus_values/values.yaml
anna-parker: ah thats ok then!

anna-parker (approved):
looks great! Would be great to still keep the nextclade links as people have found this useful but otherwise this is great, feel free to merge in and rollout :-)

loculus_values/values.yaml
corneliusroemer: done

PR #1078, opened 2026-07-21
Update default displayName creation

maverbiest:
To be merged when we roll out https://github.com/loculus-project/loculus/pull/6774

The PR incorporates the breaking change introduced in https://github.com/loculus-project/loculus/pull/6774, which now makes it so we need to add a `human_readable_pattern` if we use regex to attempt to parse the identifier field (affecting YFV and DENV)

This PR also updates the default configuration for displayName creation on PPX. By default, the displayName will be generated using `build_display_name`, which attempts to use the `specimenCollectorSampleId` in the displayName (or, if this fails, the`submissionId`, but only for direct submissions). 

Measles still uses the old displayName for now - it requires a custom displayName that we havent configured yet.

We'll need to reprocess all sequences when we roll this out to ensure the new displayName creation is used, so the PR also includes pipeline version bumps (except for measles).

🚀 Preview: https://preview-display-name-update.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -3664,7 +3664,6 @@ organisms:
       - <<: *preprocessing
         version:
           - 27
-          - 28
         replicas: 1
         configFile:
           <<: *preprocessingConfigFile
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
LGTM! Lets test this out on staging!

PR #1089, opened 2026-08-03
Clean up old pipeline versions

maverbiest:
Noticed we still had old pipeline versions in the config (left over from previous rollout).

I checked that PPX currently uses the new pipeline version for each organisms through the admin dashboard, so this PR removes these old pipeline versions.


anna-parker (approved):
LGTM! I confirmed we are on the new mpox dataset tag!

PR #1096, opened 2026-08-06
feat(config): remove older pipelines versions (except for mpox as this is currently reprocessing)

anna-parker:
…rently reprocessing)


loculus_values/values.yaml
```diff
@@ -2838,7 +2826,7 @@ organisms:
         <<: *preprocessing
         replicas: 1
         version:
-          - 27
+          - 28
```
anna-parker: wupps mistake

PR #1097, opened 2026-08-06
feat(config): temporarily revert options list changes - go back to pr…

anna-parker:
Current state:

- all organisms (except mpox) have been versioned bumped and are on the latest version - code has been merged to remove previous versions but this is not live on prod
- mpox has been bumped to v27 without alignments, v26 still exists (v26 has alignments and is without restricted options) [I attempted to bump mpox to v28 on staging but this did not bump because some sequences did not align - this has now been rolled back]

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -79,8 +79,8 @@ pipelineVersionUpgradeCheckIntervalSeconds: 300
 zstdCompressionLevel: 15
 lineageSystemDefinitions:
   mpoxOutbreakLineage:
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
maverbiest:
```suggestion
26: https://pathoplexus.github.io/silo-lineage-hierarchy-definitions/definitions/mpox/2026-07-07--14-07-11Z/outbreak-lineages.yaml
```

should we change here as well?

PR #1104, opened 2026-08-10
Add GISAID Isolate ID field to Dengue

corneliusroemer:
See https://loculus.slack.com/archives/C05F8GKR06R/p1786208915386949

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -1,4 +1,3 @@
-$schema: https://raw.githubusercontent.com/loculus-project/loculus/refs/heads/main/kubernetes/loculus/values.schema.json
 preprocessingTimeout: 600
 getSubmissionListLimitSeconds: 3600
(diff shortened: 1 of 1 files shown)


loculus_values/values.yaml
```diff
@@ -2160,6 +2161,7 @@ organisms:
           autocomplete: true
           preprocessing:
             inputs: { input: ASSIGNED_REFERENCE }
+        - <<: *gisaidIsolateIdMetadataField
```
maverbiest: CI is fialing because the anchor is defined later in the file, the definition should be up here and then mpox can use the anchor

corneliusroemer: Yep just saw, hadn't checked CI!

corneliusroemer: Moving anchor up, should work now

maverbiest (approved):
Looks good!

PR #1110, opened 2026-08-12
Narrow Yellow fever clade column to 76px

corneliusroemer:
The YFV clade field had no columnWidth set, so it fell back to the Loculus default of 130px which is wider than we need, wasting horizontal space. Most other clade/lineage fields use custom (narrower) width.


maverbiest (approved):
Thanks!

PR #1111, opened 2026-08-12
Pull YFV dataset from Nextclade Data now that it's released

corneliusroemer:
### Testing

- Adding preview to see YFV process properly just in case
- Preview looks good: 
  image

🚀 Preview: https://preview-use-nextclade-server-yfv.pathoplexus.org

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -3979,7 +3979,7 @@ organisms:
       linkOuts:
         - name: "Nextclade"
           maxNumberOfRecommendedEntries: 2500 # yellow fever is ~11kb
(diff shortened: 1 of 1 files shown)


maverbiest (approved):
Looks great to me, thanks!

I had a look at the preview as well, tested the linkout to Nextclade. I did notice that the card on the landing page shows two fewer YFV on the preview vs on prod? But maybe this is just bc it's still processing. Is there a way log in as ingest user/superuser on previews to check?

**preview**:

Image

**prod**:

Image

corneliusroemer:
There are 2 direct restricted submissions 😀

PR #1113, opened 2026-08-17
Own and shrink organism images (ebola, WNV, CCHF, Andes)

corneliusroemer:
## Why

Noticed while compressing the Zika image.

**We did not own four of the images we serve.** `ebolasudan_small.jpg`, `ebolazaire_small.jpg`, `wnv_small.jpg` and `cchf_small.jpg` come from the loculus overlay — they are in the generated `monorepo/.gitignore` (lines 980–988), so upstream controlled what pathoplexus serves.

`sync.sh` adds a path to `.gitignore` only when the file is **absent** locally, then copies with `rsync -av --ignore-existing`. So committing our own copies takes ownership permanently — no rename, no `sync.sh` change.

**Images were 359 kB of a 568 kB page load,** and these four were the only ones narrower than the card's rendered width, so the only ones upscaled by the browser.

## What changed

3:2 centre crop matching the card (`w-60 h-40` = 240×160 CSS px). Previous ratios were all different (1.11, 1.33, 1.0, 1.5) and `object-cover` cropped them to 3:2 anyway.

| image | before | after |
|---|---|---|
| `ebolasudan.jpg` | 220×198, 25.8 kB | 300×200, **19.4 kB** |
| `ebolazaire.jpg` | 220×165, 19.9 kB | 300×200, **17.5 kB** |
| `wnv.jpg` | 220×169, 18.0 kB | 256×171, **12.2 kB** |
| `cchf.jpg` | 220×165, 14.2 kB | 300×200, **11.7 kB** |
| `andv.jpg` | 276×228, 32.0 kB | 276×184, **9.4 kB** |

110 kB → 70 kB served, all now in the 10–20 kB band of the other images, none upscaled. `values.yaml` repointed off the four `_small` paths.

## Command

```bash
# the four micrographs, from upstream's high-res sources
magick IN.jpg -filter Lanczos -resize 300x -gravity center -crop 300x200+0+0 +repage \
  -strip -sampling-factor 4:2:0 -interlace JPEG -quality 58 OUT.jpg

# wnv (source only 256px) and andv (276px) — cropped at native width, no upscaling
magick wnv.jpg  -gravity center -crop 256x171+0+0 +repage \
  -strip -sampling-factor 4:2:0 -interlace JPEG -quality 58 wnv.jpg
magick andv.jpg -gravity center -crop 276x184+0+0 +repage \
  -strip -sampling-factor 4:2:0 -interlace JPEG -quality 78 andv.jpg
```

[...]

PR #1114, opened 2026-08-17
feat(actions): Claude pr reviews

maverbiest:
I basically copied over [Loculus' workflow file](https://github.com/loculus-project/loculus/blob/main/.github/workflows/claude-code-review.yml). This should work if we add a `CLAUDE_CODE_OAUTH_TOKEN` to this repository's secrets.

The one tweak I made is adding 'opened' to the array of pull request types that the PR is run on. As far as I understand this makes it so the review also runs on PRs that are opened immedately as 'ready for review', instead of only when a PR is opened as draft and then changed to 'ready for review'. Is this something we actually want (maybe also on Loculus)? If not, I will get rid of it.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,7 +2,7 @@ name: Claude Code Review
 
 on:
   pull_request:
-    types: [ready_for_review, reopened]
+    types: [opened, ready_for_review, reopened]
 
 jobs:
   review-with-tracking:
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,0 +1,75 @@
+name: Claude Code Review
+
+on:
+  pull_request:
+    types: [ready_for_review, reopened]
+
+jobs:
+  review-with-tracking:
+    if: github.event.pull_request.draft == false && github.event.pull_request.user.login != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write
+      issues: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 1
+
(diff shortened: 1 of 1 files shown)


corneliusroemer:
> As far as I understand this makes it so the review also runs on PRs that are opened immedately as 'ready for review', instead of only when a PR is opened as draft and then changed to 'ready for review'. 

Good catch! It really looks like Claude only reviews PRs that are switched from draft to ready to review. Example: https://github.com/loculus-project/loculus/pull/7082

Yes we should switch this also on Loculus then - though just thinking more about it, we often open PRs then substantially change them. The review would only happen the first time unless we add `synchronize` as additional event on which it triggers. That's what Anthropic suggests in their example workflow: https://github.com/anthropics/claude-code-action/blob/main/examples/pr-review-comprehensive.yml#L8

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: PR Review with Progress Tracking
+        uses: anthropics/claude-code-action@v1
+        with:
+          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+
+          # Enable progress tracking
+          track_progress: true
+
+          model: claude-opus-4-5-20251101
```
corneliusroemer:
```suggestion
          model: claude-opus-5
```

corneliusroemer (approved):
I've added an Oauth token created via `claude setup-token` as an organization secret so other repos can also reuse it. Let's merge and try it out :)

corneliusroemer:
It works - for now we just need to convert to draft and back to trigger the PR reviews:

https://github.com/pathoplexus/pathoplexus/pull/1113

Google Chrome 2026-08-18 16 57 03

maverbiest:
>for now we just need to convert to draft and back to trigger the PR reviews

This is not too bad I think, it makes it more of a manual thing saying you want a review rather than getting one every time you push something

PR #1119, opened 2026-08-25
fix: Replace NIAID Flickr links with archive.org links

corneliusroemer:
NIAID's flickr has been taken down apparently so our image links go 404. This PR replaces them with archive.org links.


anna-parker:
Not sure if its just me but I cant load the flickr pages or the webarchive pages

monorepo/website/src/pages/about/faq.mdx
```diff
@@ -242,15 +242,15 @@ All of the Loculus code is also open-source, and you can view it [here](https://
...
 
 The images we use from NIAID are:
-  - [CCHF](https://www.flickr.com/photos/niaid/40689899455/)
-  - [Dengue](https://www.flickr.com/photos/niaid/53786996438/) (credit shared with the CDC)
-  - [Ebola Sudan](https://www.nih.gov/sites/default/files/styles/featured_media_breakpoint-medium/public/news-events/news-releases/2023/niaid-03.jpg?itok=m4wFNGkn&timestamp=1675437186)
-  - [Ebola Zaire](https://www.flickr.com/photos/niaid/14440817981/in/album-72157646274075631/)
-  - [RSV](https://www.flickr.com/photos/niaid/52453988775/)
-  - [Measles](https://www.flickr.com/photos/niaid/53483356800)
+  - [CCHF](https://web.archive.org/web/20250201054608/https://www.flickr.com/photos/niaid/40689899455/)
+  - [Dengue](https://web.archive.org/web/20250201054608/https://www.flickr.com/photos/niaid/53786996438/) (credit shared with the CDC)
```
anna-parker, line 249: weird: 

Image

corneliusroemer-agent:
I checked all seven new URLs with curl (following redirects) and verified page content, not just status codes. **All seven return HTTP 200**, and each archived Flickr page carries the pathogen it's labelled with:

| Link | Result |
| --- | --- |
| Flickr account page | 200 — "NIAID \| Flickr", real photostream (24 photo tiles), not a login wall |
| CCHF | 200 — "Crimean-Congo Hemorrhagic Fever (CCHF) Virus" |
| Dengue | 200 — "Aedes Mosquito and Dengue Virus Particles" |
| Ebola Sudan (nih.gov) | 200, `image/jpeg`, 218 KB — serves the actual image |
| Ebola Zaire | 200 — "Ebola Virus Particles" |
| RSV | 200 — "Human Respiratory Syncytial Virus (RSV)" |
| Measles | 200 — "Measles Virus Particle" |

The embedded photos are archived too — I fetched the archived JPEGs directly and got real image bytes (CCHF 217 KB, Dengue 376 KB, Measles 332 KB, RSV 330 KB). Ebola Zaire's JPEG I couldn't confirm because archive.org started returning 429 to my IP after that burst; its page loads fine and every sibling worked, so that's throttling on my end, not a broken link.

The premise holds for Flickr: the account is `410 Gone`, each photo page `404`, and even the raw CDN asset (`live.staticflickr.com/5033/14440817981_eb676a8850_b.jpg`) now `404`s. The images are genuinely gone from Flickr, so archive.org is the only remaining source.

**@anna-parker** — your screenshot actually shows the archived page working: the title, description and "Credit: CDC and NIAID" all render. Only the photo area is blank, because Flickr lazy-loads the image via JavaScript that Wayback doesn't replay. The archived image file itself is fine — that exact Dengue JPEG fetches as 376 KB of `image/jpeg`. So the credit links resolve; if you want the picture to be visible on click, linking the archived image asset directly instead of the photo page would do it.

Two things worth flagging on my earlier inline review, which was written without network access and got these wrong:

- **Line 250 (Ebola Sudan)**: I claimed the `?itok=` query string made a Wayback capture unlikely and that wrapping it "likely turns a working link into a broken one". That's wrong — the capture exists and serves the image (200, `image/jpeg`, 218 KB). The scope observation does stand, though: this one isn't a Flickr link, and the `403` it returns is a Cloudflare "Just a moment…" bot challenge rather than a 404, so it's probably still live for real browsers. Since the title and body say "Flickr links … go 404", 1 of the 7 changed links is neither — either narrow the description or drop that hunk. That's a judgement call, not a defect.
- **Line 251 (Ebola Zaire)**: I claimed the `/in/album-.../` suffix would hit the "has not archived that URL" page. Also wrong — it returns 200 with the correct photo.

Resolving both of those threads.

Minor and non-blocking: all seven URLs reuse timestamp `20250201054608`, which is a real snapshot only for the account page; the other six silently redirect to the nearest capture (20250117, 20250411, 20250607, 20250303). That works reliably, but the timestamp isn't doing any work — `/web/2025/` would read more honestly.

I also swept the other 16 links in `faq.mdx`: all live, except the `cdc.gov` Marburg PNG (Akamai "Access Denied") and `niaid.nih.gov` (AWS WAF human verification), which are blocking automated requests rather than being dead, and are untouched by this PR. `faq.mdx` is the only file in the repo referencing Flickr, so nothing else needs changing.

The point about `.github/workflows/broken-link-check.yml` having its `pull_request`, `push` and nightly `schedule` triggers commented out still stands, and would have caught this rot.

monorepo/website/src/pages/about/faq.mdx
corneliusroemer: Yeah they didn't archive the picture 🤷 but there's still the description at least. This is the best resource for "this is where we got it from" - we could also delete the links altogether.

corneliusroemer: Ah apparently the image is backed up as well and visible via the raw resource url, just the flickr lazy load doesn't work. We could replace the link target with the raw jpg urls as you prefer

anna-parker: ah that would be nice - but will approve as its better than current state

anna-parker (approved):

PR #1121, opened 2026-08-26
fix: add header to host input field so it is correctly shown in host category on website docs

anna-parker:
Should fix the following bug as host was added without a category:
Screenshot 2026-08-26 at 14 15 32


maverbiest (approved):
Thanks!

anna-parker:
wupps this doesnt solve the problem! I will need to investigate why - maybe the header is getting dropped somehow!

PR #1124, opened 2026-08-26
Set the ingest batch size to 1000

corneliusroemer-agent:
Depends on #1123

Reduce ingest's batch size for submissions to backend from 10k to 1k to avoid timeouts (these happen frequently when all ingests run at the same time, overwhelming the backend). 1k should avoid syncs getting stuck on previews. There's no real downside to smaller bathes.

### DIFF
--- a/loculus_values/values.yaml
+++ b/loculus_values/values.yaml
@@ -3943,7 +3943,6 @@ organisms:
       configFile:
         <<: *defaultIngestConfigFile
         taxon_id: 3052505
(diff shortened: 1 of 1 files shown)


corneliusroemer-agent:
Superseded by #1129, which has the same one-line change on a branch in this repo rather than a fork, so the workflows that need repository secrets can actually run. Rebased onto current `main` now that #1123 is merged. Closing this one.

PR #1126, opened 2026-08-27
Fix dengue prM links

maverbiest:
All four dengue serotypes were downloading the prM fasta for denv1.

This PR fixes the links and adds a prepro pipeline version bump for dengue


anna-parker (approved):
Lets ensure we do very thorough testing of this on staging with db cloning beforehand, db cloning and regression testing afterwards as well as testing that mutation search as well as aligned sequence downloads are WAI.

PR #1128, opened 2026-08-27
feat(db-clone): do not perform sed replacements on protected url in version comments, add env file with rds pghost info

anna-parker:
Ignores doc file links that were affected by sed replacement during db clone.

Additionally makes the following changes/improvements to the db clone script:

- Adds an env file to add rds pghost info (was changed on the bastion but not commited, which I think is good) 

- I replaced check_command with having set -Eeuo pipefail as check_command only check the command right before it (an existing issue - we should have failed if any sed failed but this didnt happen).

ChatGPT then told me it wasnt good practice to use a pipe inside the if statement when activating pipefail as: grep -q is specifically allowed to stop reading input as soon as it finds a match: It finds prod_, so grep -q says, effectively, "I have my answer" and exits successfully with status 0. But the first grep may still be processing the file and trying to write matching lines into the pipe. Since the reader of the pipe has exited, the OS can send the first grep SIGPIPE. It then commonly exits with status 141 (128 + SIGPIPE(13)). to remedy it we can replace the pipe with one single awk statement.

Confirmed with local testing that the diff is gone and regression tests showed clone worked correctly:
image

### DIFF
--- a/scripts/db-clone/clone-prod-to-staging.sh
+++ b/scripts/db-clone/clone-prod-to-staging.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -Eeuo pipefail
+
 # This script orchestrates the cloning of Keycloak and Loculus databases from production to staging
 # Keycloak is dumped second and loaded first to prevent potential race conditions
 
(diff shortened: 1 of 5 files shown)


scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -14,50 +16,46 @@ STAGING_LOC_DB="pathoplexus_staging_loculus"
...
-}
-
 # Note: Could screw up columns and values that contain `prod` etc
 # For now not an issue but might eventually want to be more surgical
 perform_sed_replacements() {
     local file="$1"
     echo "Performing sed replacements on $file..."
     # Abort the script if a line with `@` contains the word "prod" to prevent altering sensitive data
-    if grep '@' "$file" | grep -q 'prod_'; then
+    if awk '/@/ && /prod_/ { found=1; exit } END { exit !found }' "$file"; then
```
theosanderson, line 25: I don't fully understand this? Could you explain the rationale? But nbd

theosanderson:
@codex review

theosanderson (approved):

scripts/db-clone/clone-prod-to-staging.sh
anna-parker: sorry I meant to add to the PR description, I replaced `check_command` with having `set -Eeuo pipefail` as `check_command` only check the command right before it (an existing issue - we should have failed if any sed failed but this didnt happen).

ChatGPT then told me it wasnt good practice to use a pipe inside the if statement when activating pipefail as: `grep -q is specifically allowed to stop reading input as soon as it finds a match: It finds prod_, so grep -q says, effectively, "I have my answer" and exits successfully with status 0. But the first grep may still be processing the file and trying to write matching lines into the pipe. Since the reader of the pipe has exited, the OS can send the first grep SIGPIPE. It then commonly exits with status 141 (128 + SIGPIPE(13)).` to remedy it we can replace the pipe with one single awk statement.

theosanderson: kk, thanks

PR #1131, opened 2026-08-28
Clean up old dengue pipeline version

maverbiest:
Cleaning up from yesterday's rollout


anna-parker (approved):
thanks!

PR #1142, opened 2026-09-08
feat: enable files on PPX, add embl annotation files, reprocess everything

### LINKED ISSUE #6695: MVP: Replicating files to staging for rollout testing
Thinking a bit more about file sharing I realized files make it much harder to replicate production state to staging for testing.

Right now to replicate prod state on staging we copy the 2 databases (keycloak and loculus) and do a bit of string replacement for the few differences between staging and prod - like the db user name.

As soon as files are added, there's extra state that's not contained in the 2 dbs. If we carry on as is staging will point at prod s3. If staging doesn't get read and write rights on prod s3 files, prepro will crash. But we can't just give staging those rights - otherwise garbage collection from staging will delete new prod files and vice versa.

So as part of replication from prod -> staging we'll have to copy the whole s3 data used by prod over to staging. Staging will need its own bucket as backend uses hardcoded toplevel `/files/`.

Syncing should hopefully work with something like:

```sh
aws s3 sync s3://source-bucket/files/ s3://dest-bucket/files/ \
    --delete \
    --exact-timestamps
```

which we'll need to add here: https://github.com/pathoplexus/pathoplexus/blob/main/scripts/db-clone/clone-prod-to-staging.sh

Discussed in https://loculus.slack.com/archives/C05G172HL6L/p1781622089643369

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/6695, https://github.com/loculus-project/loculus/issues/6772

This enables files on this preview, adding an AWS S3 bucket for staging and production.
It also enables embl annotations files, note on dengue uses the higher level organism name still - this can be corrected later.

In order to add annotation files all data needs to be reprocessed.

- [x] create S3 buckets, IAM users and groups and added to sealed secrets
- [x] confirmed annotations and raw read file submission/download paths work
- [x] set CORS policies on staging bucket
- [x] confirmed `aws sync` works between buckets with minimal permissions
- [x] add `aws sync` and AWS IAM user configuration docs to db clone
- [x] set up life cycle policies: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html
- [x] investigate backups/versioning: https://github.com/loculus-project/loculus/issues/7321
- [x] configure and test db clone on bastion - db clone worked and a test file was transfered from prod to staging!


loculus_values/values.yaml
```diff
@@ -11,9 +11,18 @@ gitHubIssuesUrl: https://github.com/pathoplexus/pathoplexus/issues
 issuesEmail: bug@pathoplexus.org
 runDevelopmentKeycloakDatabase: true
 runDevelopmentS3: false
```
anna-parker, line 13: when we actually merge
```suggestion
runDevelopmentS3: true
```

loculus_values/values.yaml
```diff
@@ -23,6 +32,12 @@ dataUseTerms:
     open: https://pathoplexus.org/about/terms-of-use/open-data
     restricted: https://pathoplexus.org/about/terms-of-use/restricted-data
 secrets:
+  s3-bucket:
```
anna-parker, line 35: add to staging when actually merging

loculus_values/values.yaml
```diff
@@ -11,9 +11,18 @@ gitHubIssuesUrl: https://github.com/pathoplexus/pathoplexus/issues
 issuesEmail: bug@pathoplexus.org
 runDevelopmentKeycloakDatabase: true
 runDevelopmentS3: false
+createTestAccounts: true
```
tombch: This needs reverting before merge

loculus_values/environment_specific_values/production.yaml
```diff
@@ -73,6 +79,17 @@ seqSets:
     writeEnabled: true
 podScheduling:
   podPriorityValue: 2000
+runDevelopmentS3: false
+s3:
+  enabled: true
+  gcEnabled: true
+  gcFrequencyMinutes: 720
+  gcInitialDelayMinutes: 15
+  gcOrphanRetentionPeriodMinutes: 1440
```
anna-parker, line 88: maybe lets wait 3 days?

loculus_values/environment_specific_values/staging.yaml
```diff
@@ -62,8 +62,25 @@ secrets:
...
+      secretKey: AgB8b135FsMHqko+wQuD2lWUtuaRuG42n/AgtSU7vJbAvoner02QG3vxm4TQVuxrNjmN2oRCj8Go9+Wai9eL5brde69m/O2zGpJpQik1YKmdpQKiLEcOALY4u5C7iN072en8R7hjYqC0HOYH5pek8sO8OnZtCGbhCiEg6UpbetS8uEoLWRVcLwcBlGN+sxypl0R6uuL9ZOkv6+WD5ZauDoI9lag4NItW8YwQZdOumUn5yfJNc3vuhJaE6qHuw0ufUn9isvwjBd12rfqiFTKOzdWbwesFqxONTyY8Fe7G4A+PIFbJEkP0pIVeJijvkfz5g3rI4vCJwBFT1AXVGllZVZDfoI2cii0K1zyhKTcvV6V6RqjUa2SRAMc5pW98fcYw2wqDpsWR5QpC4LPwlo0npMfriBWbURHTf3hs9c1Q38D9CEZYdDSJ7pvSfwikZUuJ3781+0rBXQP2IU5ozyAbRcI5S9niy9elIbyaXQTMripVrNNmzWLt4rIuhZnjqZeEskAwxkk5+jlYW9PvMfyTsfbkoUcmtyHVheF0VFr1VnMKKXe5P8QfVKwUAWLXHImPoGc4cApvmGmgs345X+dt5au8Zli4+09hKnTz/u1t+pUuh5RIf7i41xkpHWg8PSpYm4YKVBBxZywFFz8s/nzQSkJCCgO+I5rxRWK9MoCC65h6I9vXdbtDjrikFCAGFLozADM39kwrg0HQjrPfXY8WY4eMFX0TcDgRuIUP/4t5faemxyawkOzH3DFL
 podScheduling:
   podPriorityValue: 1000
+runDevelopmentS3: false
+s3:
+  enabled: true
+  gcEnabled: true
```
anna-parker:
```suggestion
  gcEnabled: false
```

loculus_values/environment_specific_values/production.yaml
```diff
@@ -73,6 +79,17 @@ seqSets:
     writeEnabled: true
 podScheduling:
   podPriorityValue: 2000
+runDevelopmentS3: false
+s3:
+  enabled: true
+  gcEnabled: true
```
anna-parker:
```suggestion
  gcEnabled: false
```

loculus_values/values.yaml
```diff
@@ -2509,7 +2554,7 @@ organisms:
           - 8
         configFile:
           <<: *preprocessingConfigFile
-          create_embl_file: false
+          taxon_id: 1980456
           scientific_name: "Andes Virus"
```
anna-parker: oh crap this is wrong... I think claude added andes so that kinda makes sense - we need to also change for ENA submission
```suggestion
          scientific_name: "Orthohantavirus andesense"
```

loculus_values/values.yaml
```diff
@@ -4063,6 +4129,9 @@ organisms:
           - 31
         configFile:
           <<: *preprocessingConfigFile
+          taxon_id: 11089
+          scientific_name: "Yellow Fever Virus"
```
anna-parker: probably not too important
```suggestion
          scientific_name: "Yellow fever virus"
```

loculus_values/values.yaml
```diff
@@ -4251,3 +4310,9 @@ resources:
       cpu: "100m"
     limits:
       memory: "3Gi" # Backend requires at least 635741K of memory
+  raw-reads-processing:
+    requests:
+      memory: "8Gi" # deacon loads the whole panhuman index into memory (~3.3Gi on disk, ~4.8Gi peak while loading)
+      cpu: "50m"
+    limits:
+      memory: "8Gi"
```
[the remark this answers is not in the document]
anna-parker: it shouldnt enable raw read submission yet... so maybe I messed sth up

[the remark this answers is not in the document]
anna-parker: ah maybe just that these resource values are here - I can delete them

scripts/db-clone/clone-prod-to-staging.sh
```diff
@@ -42,13 +49,26 @@ perform_sed_replacements() {
...
 
 echo "Dumping production Keycloak database..."
 $CHILD_SCRIPT dump $PROD_KC_DB $PROD_KC_DUMP
 perform_sed_replacements $PROD_KC_DUMP
 
+# Sync files after the clone so no files referenced by the db can be missing (there might be
+# additional unreferenced files in the bucket but that is ok, to prevent additional files we
+# could enable s3 versioning and pick whichever version has LastModified <= T
+echo "Syncing S3 buckets..."
+sync_s3_buckets
```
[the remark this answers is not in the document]
anna-parker: ah yes I meant dumps - will update

loculus_values/environment_specific_values/production.yaml
```diff
@@ -73,6 +79,17 @@ seqSets:
     writeEnabled: true
 podScheduling:
   podPriorityValue: 2000
+runDevelopmentS3: false
+s3:
+  enabled: true
+  gcEnabled: false
+  gcFrequencyMinutes: 720
+  gcInitialDelayMinutes: 15
+  gcOrphanRetentionPeriodMinutes: 1440
```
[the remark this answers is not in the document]
anna-parker: fine for now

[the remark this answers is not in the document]
corneliusroemer: What's your reason for keeping it off? Worry that it might do something bad and want to test on staging when we have some data in prod?

loculus_values/values.yaml
```diff
@@ -10,10 +10,16 @@ gitHubMainUrl: https://github.com/pathoplexus/pathoplexus
 gitHubIssuesUrl: https://github.com/pathoplexus/pathoplexus/issues
 issuesEmail: bug@pathoplexus.org
 runDevelopmentKeycloakDatabase: true
-runDevelopmentS3: false
+runDevelopmentS3: true
```
corneliusroemer, line 13: If possible might be nice to use actual AWS s3 also for previews not just for staging/prod.

IIUC, to be able to reuse a bucket we'd need to be able to use a path-prefix not put `/files` only at the root, though. And current Loculus apparently doesn't support that. Might be worth doing at some point to make previews more prod/staging-like.

loculus_values/values.yaml
```diff
@@ -1773,6 +1790,7 @@ defaultOrganismConfig: &defaultOrganismConfig
       configFile: &preprocessingConfigFile
         log_level: DEBUG
         batch_size: 100
+        create_embl_file: true
```
corneliusroemer, line 1793: In theory we should probably only change this line for the new processing versions? It doesn't really matter probably as only new files would be affected - but in principle we only change prepro config when reprocessing. This could matter if the new config/version crashes somehow - then we still have the old prior version functioning.

anna-parker: I think this is fine, will only mean that if someone submits while we are reprocessing the embl file will show up for that submission before the others (might even be a good way to ensure there are no issues with files for different pipeline versions)

loculus_values/values.yaml
anna-parker: Hmmm... when we delete a preview I think the files will stay in the AWS bucket? we would need to configure some policies there, so think its best to hold off for now

loculus_values/values.yaml
```diff
@@ -1838,8 +1856,12 @@ organisms:
         replicas: 1
         version:
           - 31
+          - 32
         configFile:
           <<: *preprocessingConfigFile
+          taxon_id: 186538
+          scientific_name: "Zaire ebolavirus"
+          molecule_type: "genomic RNA"
```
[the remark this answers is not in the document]
anna-parker: ah yes much nicer, was a larger change than anticipated also 
```
<<: *preprocessingConfigFile
  <<: &ebolaZaireTaxon
```
is actually not allowed it needs to be restructured to have
```
<<: [*preprocessingConfigFile, *ebolaZaireTaxon]

File: scripts/db-clone/clone-prod-to-staging.sh [hunk 15]
```diff
@@ -42,13 +44,25 @@ perform_sed_replacements() {
...
 echo "Dumping production Loculus database..."
 $CHILD_SCRIPT dump $PROD_LOC_DB $PROD_LOC_DUMP
 
 echo "Dumping production Keycloak database..."
 $CHILD_SCRIPT dump $PROD_KC_DB $PROD_KC_DUMP
 perform_sed_replacements $PROD_KC_DUMP
 
+# Sync files after the dump so no files referenced by the dumped db can be missing (there might
+# be additional unreferenced files in the bucket but that is ok, to prevent additional files we
+# could enable s3 versioning and pick whichever version has LastModified <= T
```
corneliusroemer, line 62: One (unlikely/rare) race could be when someone completes a multi-part file after we synced. To make this less likely we should probably do the sed replacements _after_ we sync s3. Just to minimize the time window.

We could theoretically also print something like the last modification time of s3 files before the db dump starts, and after the s3 sync finished and erroring when the 2 differ.

corneliusroemer: Ah wait I see, you're already doing sync before sed replacements - just not the kc sed - that could also move after here

anna-parker: Im not sure if we gain much for that as i dont think the KC db has any links to files? but I might be wrong

loculus_values/values.yaml
corneliusroemer: Yes, they would stay but cost so little that we can just delete it once a year or have a hook that deletes, but probably not worth it for a long time.

scripts/db-clone/clone-prod-to-staging.sh
anna-parker: true... we could get the most recent `LastModified` for all files in the bucket and abort if sth changed in the time it took us to dump the db:
```
aws s3api list-objects-v2 \
  --bucket my-bucket \
  --query 'max_by(Contents, &LastModified).{Key: Key, LastModified: LastModified}'
```
however, claude thinks this might get slow when we have a lot of objects... what do you think?

loculus_values/environment_specific_values/production.yaml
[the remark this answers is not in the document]
anna-parker: @theosanderson really wanted us to not release with GC on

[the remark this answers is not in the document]
anna-parker: I personaly think its safe but we can hold off for a couple months, alternatively I think @tombch and I came up with a good solution that should allow us to enable it faster: https://github.com/loculus-project/loculus/issues/7321

loculus_values/values.yaml
corneliusroemer: I'm just estimating costs of doing that for previews - and also in general. I haven't seen a cost analysis for something like annotations.

First guess is that the writes of 300k files is the most expensive part short term - roughly $1-2. But there might be more charges, looking into it right now.

anna-parker: we would need to make a new bucket for each preview though I think?

corneliusroemer (approved):
Looks good, let's test thoroughly on staging!

scripts/db-clone/clone-prod-to-staging.sh
corneliusroemer: > Ah wait I see, you're already doing sync before sed replacements - just not the kc sed - that could also move after here

I meant that we would shorten the time window during which the race could happen at no cost as sed is deterministic offline thing that can be run any time.

corneliusroemer: > true... we could get the most recent LastModified for all files in the bucket and abort if sth changed in the time it took us to dump the db:

> aws s3api list-objects-v2 \
  --bucket my-bucket \
  --query 'max_by(Contents, &LastModified).{Key: Key, LastModified: LastModified}'
however, claude thinks this might get slow when we have a lot of objects... what do you think?

I'd query the db directly rather than s3? I hope we have a timestamp of sorts in there 😆

anna-parker: I will skip this for now

PR #1152, opened 2026-09-18
fix(ci): pin deacon index

anna-parker:
counterpart to https://github.com/loculus-project/loculus/pull/7355 - Pin deacon index to the current index, produced from https://github.com/loculus-project/loculus/pull/7196 (this is to ensure if someone reruns the job the index we use on production cannot be changed)


corneliusroemer (approved):
Tested that the image can be pulled with this tag

